PR #17, opened 2020-06-22
Gitpod Setup

MauroGuerrero20:
### Features:
- Setup Gitpod for project. (All configuration files should be ready to go)

PR #21, opened 2020-06-23
Implements BookList Object

### LINKED ISSUE #5: Implement Book Lists Object
## Book Lists:

- Stores User's id
- Stores User's Bookshelf id (Used to get bookshelf from Google Books API)
- Stores list of collaborators (Use to implement sharing feature)
- Stores Unique Book List id

TODO: https://github.com/STEP-Pod-Capstone-Project/Capstone-Project/issues/22#issue-644216068
- Integrate Datastore or Firebase with the BookList Object.

MauroGuerrero20:
## Features
- Implements BookList Object on Java Backend
- Purpose of object is to store:
  - User's ID
  - User's Bookshelf ID (Use to get a Bookshelf using Google Books API)
  - List of Collaborators (Use to share BookList)
  - Unique BookList ID

#### Depends on https://github.com/STEP-Pod-Capstone-Project/Capstone-Project/pull/17#issue-438143543


capstone-backend/src/main/java/com/google/sps/data/BookList.java
```diff
@@ -0,0 +1,73 @@
...
+ * 
+ * Additionally, the Book List object also stores collaboratorsIDs.
+ * 
+ * The 'collaboratorsIDs' variable purpose is to allow the implemation of a
+ * sharing feature between Book Lists.
+ * 
+ */
+public final class BookList {
+
+  private final long id = new Random().nextLong();
```
steven-solar: Maybe we should be able to input this value for now, as a way to test.

capstone-backend/src/main/java/com/google/sps/data/BookList.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * The Book List object stores userID and bookshelfID.
+ * 
+ * These IDs variables are meant to be fetch from the Frontend, to extract a
+ * User's data using the Google Books API.
+ * 
+ * Additionally, the Book List object also stores collaboratorsIDs.
+ * 
+ * The 'collaboratorsIDs' variable purpose is to allow the implemation of a
```
antoniolinhart: nit: implementation

capstone-backend/src/main/java/com/google/sps/data/BookList.java
```diff
@@ -0,0 +1,73 @@
...
+ * sharing feature between Book Lists.
+ * 
+ */
+public final class BookList {
+
+  private final long id = new Random().nextLong();
+  private String bookshelfID = "";
```
steven-solar: up here I typically don't assign my variables (eg. private String bookshelfID; and private ArrayList<Long> collaboratorIDs;) and then assign them only in the constructor.

capstone-backend/src/main/java/com/google/sps/data/BookList.java
```diff
@@ -0,0 +1,73 @@
...
+
+  public void setBookshelf(String bookshelfID) {
+    this.bookshelfID = bookshelfID;
+  }
+
+  public void setUserID(Long userID) {
+    this.userID = userID;
+  }
+
+  public void setCollaboratorsIDs(ArrayList<Long> collaboratorsIDs) {
```
steven-solar: Maybe create an add/remove collaborator function. Think we'd need those in practice.

capstone-backend/src/main/java/com/google/sps/data/BookList.java
antoniolinhart: Should we make this a constructor variable we can pass in like @steven-solar did in his PR since we don't have a method to store and create these ids yet?

capstone-backend/src/test/java/com/google/sps/BookListTest.java
```diff
@@ -0,0 +1,88 @@
...
+    assertNull(bookList1.getUserID());
+    assertEquals("", bookList1.getBookshelfID());
+    assertEquals(true, bookList1.getCollaboratorsIDs().isEmpty());
+
+    assertNotNull(bookList1.getId());
+
+    bookList2.clear();
+    assertEquals(bookList1.isEmpty(), bookList2.isEmpty());
+  }
+}
```
steven-solar, line 126: once you create addCollaborator() and removeCollaborator() functions, would be nice to test those here as well.

steven-solar (changes requested):
Looks good! After a few changes should be good to go.

capstone-backend/src/main/java/com/google/sps/data/BookList.java
```diff
@@ -0,0 +1,73 @@
...
+    this.collaboratorsIDs = collaboratorsIDs;
+  }
```
antoniolinhart: It would be helpful to have a addCollaboratorById() and removeCollaboratorById() to enable the user to edit the list of collaborators

capstone-backend/src/test/java/com/google/sps/BookListTest.java
```diff
@@ -0,0 +1,88 @@
...
+    assertEquals(COLLABORATORS_IDs_1, bookList1.getCollaboratorsIDs());
+
+    assertEquals(USER_ID_2, bookList2.getUserID());
+    assertEquals(BOOKSHELF_ID_2, bookList2.getBookshelfID());
+    assertEquals(true, bookList2.getCollaboratorsIDs().isEmpty());
+  }
+
+  @Test
+  public void testSetters() {
+
```
antoniolinhart, line 80: Some of your tests have a line between the function declaration and code and others don't, it's good to keep it consistent

capstone-backend/src/test/java/com/google/sps/BookListTest.java
```diff
@@ -0,0 +1,88 @@
...
+    assertEquals(true, bookList1.isEmpty());
+    assertNull(bookList1.getUserID());
+    assertEquals("", bookList1.getBookshelfID());
+    assertEquals(true, bookList1.getCollaboratorsIDs().isEmpty());
+
+    assertNotNull(bookList1.getId());
+
+    bookList2.clear();
+    assertEquals(bookList1.isEmpty(), bookList2.isEmpty());
+  }
```
antoniolinhart, line 125: It appears that this test relies on being the last one to run in the test set, if this gets run before another it could cause your `get` tests to fail. You could fix this by creating two new objects here or making your @Before a beforeeach

antoniolinhart (changes requested):
Just a few small changes, but looks good otherwise!

capstone-backend/src/main/java/com/google/sps/data/BookList.java
MauroGuerrero20: Done.

capstone-backend/src/main/java/com/google/sps/data/BookList.java
MauroGuerrero20: Ok for now I'll add a setter.

capstone-backend/src/main/java/com/google/sps/data/BookList.java
MauroGuerrero20: Will implement.

capstone-backend/src/main/java/com/google/sps/data/BookList.java
MauroGuerrero20: Ok

capstone-backend/src/main/java/com/google/sps/data/BookList.java
MauroGuerrero20: Ok

capstone-backend/src/test/java/com/google/sps/BookListTest.java
MauroGuerrero20: Hi Antonio, if I add the ```@before``` to the test it still requires the creation of two new objects. I initially design the test to depend on each other, in order to test just two BookList objects, instead of creating two new objects for each test.

capstone-backend/src/test/java/com/google/sps/BookListTest.java
MauroGuerrero20: Ok

capstone-backend/src/test/java/com/google/sps/BookListTest.java
MauroGuerrero20: ok

capstone-backend/src/main/java/com/google/sps/data/BookList.java
MauroGuerrero20: Done.

capstone-backend/src/test/java/com/google/sps/BookListTest.java
antoniolinhart: Does your code still work if you just make the @Before tag a @BeforeEach instead?

That way it'll run the code before each test and not just once. Generally tests shouldn't be coupled together or rely on the order they are run

antoniolinhart: Edit: Found out that `BeforeEach` is in JUnit 5 and was `Before` in JUnit 4. Your tests should work fine, I was just looking at the wrong version!

antoniolinhart (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/data/BookList.java
```diff
@@ -17,8 +17,8 @@
 public final class BookList {
 
   private Long id;
-  private String bookshelfID;
   private Long userID;
+  private String bookshelfID = "";
   private ArrayList<Long> collaboratorsIDs;
```
steven-solar: I think it might be better to make this a HashSet rather than an ArrayList. We should also discuss if we want to make everything implement the Collection interface, so we have consistent style.

capstone-backend/src/main/java/com/google/sps/data/BookList.java
```diff
@@ -74,22 +74,30 @@ public String getBookshelfID() {
     return collaboratorsIDs;
   }
 
-  public void addCollaborator(Long collaborator){
-    collaboratorsIDs.add(collaborator);
+  public Long getCollaborator(int index){
+    return collaboratorsIDs.get(index);
```
steven-solar: Similarly, this should be changed to work for a Set rather than a List.

capstone-backend/src/main/java/com/google/sps/data/BookList.java
```diff
@@ -74,22 +74,30 @@ public String getBookshelfID() {
...
-    collaboratorsIDs.add(index, collaborator);
+  public void addCollaborator(Long collaboratorID){
+    collaboratorsIDs.add(collaboratorID);
   }
 
-  public boolean removeCollaborator(Long collaborator){
-    return collaboratorsIDs.remove(collaborator);
+  public void addCollaborator(int index, Long collaboratorID){
+    collaboratorsIDs.add(index, collaboratorID);
+  }
```
steven-solar: Same here.

steven-solar (approved):

geoffbass (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/data/BookList.java
steven-solar: Or should just be removed.

PR #26, opened 2020-06-26
Fetch clubs from backend in MyClubs

### LINKED ISSUE #7: Implement Book Clubs
### Features

User will be able to:

- Create a club 

- Add members (by username/email)

- Set the Book Id

- Set the Announcement (String)

- Push Announcement to email

steven-solar:
This PR fetches clubs from the backend to render on the MyClubs.js component. Additionally, it changes non-built in react functions (in the Club related components) to utilize ES6 arrow notation, so that we do not need to bind functions in the constructor.  We decided as a team to use arrow notation for non-built in functions, and use the classic function syntax for functions built in to react (eg. componentDidMount()). This PR also removes an uneeded jquery dependency created in the last PR.


capstone-frontend/src/components/CreateClub.js
```diff
@@ -3,12 +3,7 @@ import { withRouter } from 'react-router-dom';
 import '../styles/Clubs.css';
 
 class CreateClub extends Component {
-  constructor(props) {
-    super(props);
-    this.handleSubmit = this.handleSubmit.bind(this);
-  }
-
-  handleSubmit(e) {
+  handleSubmit = (e) => {
```
antoniolinhart: Nice use of an arrow function here!

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -90,7 +90,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       clubs.add(club);
     }
     Gson gson = new Gson();
-    response.setContentType("application/json;");
+    response.setContentType("text/json;");
```
antoniolinhart: I think I may have told you wrongly when I said to use text/json instead of application/json. It looks like application/json is the official MIME media type. I think they have the same functionality though!
https://stackoverflow.com/a/477819

antoniolinhart (approved):
LGTM!

capstone-frontend/src/components/CreateClub.js
steven-solar: Thanks!

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: Oh interesting, I'll change it back.

capstone-frontend/src/components/MyClubs.js
```diff
@@ -34,13 +25,15 @@ class MyClubs extends Component {
 
   render() {
     let clubArray = [];
-    this.state.clubs.forEach(c => {
+    [...this.state.clubs].forEach(c => {
```
VivekShankar: Was there a specific reason you wanted to use the spread operator here vs. the original code?

VivekShankar (approved):
LGTM

MauroGuerrero20 (approved):
LGTM. Just remember to tested if it deploy correctly.

MauroGuerrero20 (changes requested):
Forgot to mention to convert .js files to .jsx.

capstone-frontend/src/components/MyClubs.js
steven-solar: At some point I was getting an error without it, but it seems that was related to the backend response. Changing back to the ```.forEach()``` now. Thanks!

MauroGuerrero20 (approved):
LGTM

PR #28, opened 2020-06-29
Firestore setup

### LINKED ISSUE #22: Add Firebase integration with BookList Object

steven-solar:
This PR sets up a connection to Google Firestore on the backend, and uses this to create a servlet for Books to serve as an example for future work. For now, the Firestore does not require special credentials, which will be set up at a later date. The helper class Utility.java is created, which has a public static getter method to retrieve an instance of the Firestore database for use in servlets. It also has a public static method to turn request bodies into JsonObjects. This will trim down duplicate code in future work, as setup is now only one line, as is turning requests into Jsons. 

As Google Firestore uses Strings for IDs, this PR also changes ids to be Strings rather than longs in the User and Book class, and changes tests accordingly. Firestore also requires Objects to have a default empty constructor, which is added to the User and Book class in this PR.


capstone-backend/src/main/java/com/google/sps/servlets/BookServlet.java
```diff
@@ -0,0 +1,172 @@
...
+
+    Gson gson = new Gson();
+    response.setContentType("application/json;");
+    response.getWriter().println(gson.toJson(bookList));
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    JsonObject jsonObject = createRequestBodyJson(request);
+    System.out.println(jsonObject);
```
antoniolinhart: nit: remove println

capstone-backend/src/main/java/com/google/sps/servlets/BookServlet.java
```diff
@@ -0,0 +1,172 @@
...
+        List<QueryDocumentSnapshot> documents = future.get().getDocuments();
+        for (QueryDocumentSnapshot document : documents) {
+          bookList.add(document.toObject(Book.class));
+        }
+      } catch (Exception e) {
+        System.out.println("Error: " + e);
+      }
+    }
+
+    Gson gson = new Gson();
```
antoniolinhart: Applies to the post as well - since it is used in multiple places, it might be helpful to have it as a field variable for the class

capstone-backend/firebase-debug.log
```diff
@@ -0,0 +1,16 @@
...
+[debug] [2020-06-24T21:01:32.444Z] ----------------------------------------------------------------------
+[debug] [2020-06-24T21:01:32.444Z] 
+[info] i  Firebase optionally collects CLI usage and error reporting information to help improve our products. Data is collected in accordance with Google's privacy policy (https://policies.google.com/privacy) and is not used to identify you.
+ 
+[info] i  To change your data collection preference at any time, run `firebase logout` and log in again. 
+[info] 
+[info] Visit this URL on this device to log in:
+[info] https://accounts.google.com/o/oauth2/auth?client_id=563584335869-fgrhgmd47bqnekij5i8b5pr03ho849e6.apps.googleusercontent.com&scope=email%20openid%20https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcloudplatformprojects.readonly%20https%3A%2F%2Fwww.googleapis.com%2Fauth%2Ffirebase%20https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fcloud-platform&response_type=code&state=821269329&redirect_uri=http%3A%2F%2Flocalhost%3A9005
+[info] 
+[info] Waiting for authentication...
```
antoniolinhart: Should we remove these logs so they aren't in our repo?

capstone-backend/src/test/java/com/google/sps/UserTest.java
```diff
@@ -59,19 +65,19 @@ public void testUsername() {
 
   @Test 
   public void testAddFriend() {
-    userOne.addFriend(new Long(3));
+    userOne.addFriend(ID_THREE);
     Assert.assertEquals(userOne.getFriends().size(), 1);
-    for (Long l : userOne.getFriends()) {
-      Assert.assertEquals(l, new Long(3));
+    for (String l : userOne.getFriends()) {
```
antoniolinhart: Not super important, but maybe you could make it String friendId instead of String l

capstone-backend/firebase-debug.log
```diff
@@ -0,0 +1,16 @@
...
+[debug] [2020-06-24T21:01:32.443Z] CLI Version:   8.4.3
+[debug] [2020-06-24T21:01:32.443Z] Platform:      linux
+[debug] [2020-06-24T21:01:32.444Z] Node Version:  v12.16.1
+[debug] [2020-06-24T21:01:32.444Z] Time:          Wed Jun 24 2020 21:01:32 GMT+0000 (Coordinated Universal Time)
+[debug] [2020-06-24T21:01:32.444Z] ----------------------------------------------------------------------
+[debug] [2020-06-24T21:01:32.444Z] 
+[info] i  Firebase optionally collects CLI usage and error reporting information to help improve our products. Data is collected in accordance with Google's privacy policy (https://policies.google.com/privacy) and is not used to identify you.
+ 
+[info] i  To change your data collection preference at any time, run `firebase logout` and log in again. 
+[info]
```
MauroGuerrero20: gitignore this file?

capstone-backend/src/main/java/com/google/sps/servlets/BookServlet.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+    if (jsonObject.get("review") != null) {
+      review = jsonObject.get("review").getAsString();
+    }
+    Book book = new Book(id, userID, gbookID, hasRead, rating, review);
+    ApiFuture<WriteResult> future = books.document(id).set(book);
+    try {
+      System.out.println("Update time : " + future.get().getUpdateTime());
+    } catch (Exception e) {
+      System.out.println("Error: " + e);
```
antoniolinhart: nit: System.err.println

capstone-backend/src/main/java/com/google/sps/servlets/BookServlet.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+    if (jsonObject.get("review") != null) {
+      update.put("review", jsonObject.get("review").getAsString());
+    }
+
+    ApiFuture<WriteResult> writeResult = books.document(id).set(update, SetOptions.merge());
+    try {
+      System.out.println("Update time : " + writeResult.get().getUpdateTime());
+    } catch (Exception e) {
+      System.out.println("Error: " + e);
```
antoniolinhart: Same above

antoniolinhart (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/servlets/BookServlet.java
steven-solar: Good idea, will add to constructor

capstone-backend/src/main/java/com/google/sps/servlets/BookServlet.java
steven-solar: Good catch, going through and changing all of them.

capstone-backend/src/test/java/com/google/sps/UserTest.java
steven-solar: Sure thing, forgot to change the "l" when I changed from Long to String.

capstone-backend/firebase-debug.log
steven-solar: good catch, removed it.

MauroGuerrero20 (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/servlets/BookServlet.java
```diff
@@ -0,0 +1,138 @@
...
+    db = Utility.getFirestoreDb();
+    books = db.collection("books");
+    gson = new Gson();
+  }
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    if (request.getParameter("id") != null) {
+      DocumentReference docRef = books.document(request.getParameter("id"));
+      ApiFuture<DocumentSnapshot> future = docRef.get();
```
VivekShankar, line 53: Have you tried testing (locally is fine) what happens when you make a GET request for some "id" that doesn't exist in the database (or if the database is empty)? Do you get the `"Error: no such document!"` case?

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.gson.JsonParser;
+
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+import java.io.IOException;
+
+import javax.servlet.http.HttpServletRequest;
+
+
+public class Utility {
```
VivekShankar, line 28: Consider making specific utilities for different purposes (e.g. `DatabaseUtility` holding database-related methods) in the future if this class grows.

VivekShankar (approved):
LGTM - nice work Steven!

capstone-backend/src/main/java/com/google/sps/servlets/BookServlet.java
steven-solar: Yes you do in that case. In the documentation for Firestore it says "if there is no document at the location referenced by docRef, the resulting document will be empty and calling exists on it will return false," (check it out here: https://firebase.google.com/docs/firestore/query-data/get-data#get_a_document) which is what I based this on. 

I figured the best way to handle this would just be printing to System.err and returning a null book, but can change that up if there's a better way.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: That's a good idea, and will definitely think about breaking this class apart if it gets unwieldy.

capstone-backend/firebase-debug.log
steven-solar: Just removed it altogether.

PR #32, opened 2020-07-01
Club flow: firebase

### LINKED ISSUE #7: Implement Book Clubs

steven-solar:
This PR re-creates backend functionality to get either all clubs or a club by its id, to post a club and to update a club via a put request. Before, this functionality was implemented using Google Cloud Datastore. This PR creates this same functionality (and adds filtering by ID) in Firebase. For now, this only supports basic club fields (eg. not memberIDs, inviteIDs or posts, all of which are Lists). 

The Collections and Sets in the Club.java class had to be changed to Lists as Firebase does not support Collections or Sets (https://github.com/firebase/firebase-admin-java/issues/78). Firebase essentially only supports storing maps, but can store Lists by mapping indices to their values. For example it would store Arrays.asList("foo", "bar") as:
```
{
  "1": "foo"
  "2": "bar"
}
```
However, since Collections and Sets are inherently unordered, it cannot serialize these the way it can Lists. This is explained in the above linked issue from the Firebase repository. Additionally, IDs were changed from Longs to Strings for Firebase support. The appropriate testing was similarly updated. 

The ClubPage component now performs its own fetch to get club data, using the ID to get a Club from the backend. It displays this data in a very basic form (see screenshot below for an example using some test data).

Screenshot 2020-07-01 at 12 54 05 AM - Display 1

### DIFF
--- a/capstone-backend/src/main/java/com/google/sps/data/Club.java
+++ b/capstone-backend/src/main/java/com/google/sps/data/Club.java
@@ -1,29 +1,23 @@
 package com.google.sps.data;
 
-import java.util.Collection;
 import java.util.ArrayList;
-import java.util.HashSet;
+import java.util.List;
 
- 
-/**
- * Class representing a Club.
- *
- * Note: The private variables in this class are converted into JSON.
- */
+
 public class Club {
   
-  private final long id;
+  private final String id;
   private String name;
   private String description;
   private String announcement;
-  private Collection<String> posts;
-  private long ownerID;
-  private Collection<Long> memberIDs;
-  private Collection<Long> inviteIDs;
+  private List<String> posts;
+  private String ownerID;
+  private List<String> memberIDs;
+  private List<String> inviteIDs;
   private String gbookID;
 
-  public Club(long id, String name, String description, String announcement, Collection<String> posts, 
-      long ownerID, Collection<Long> memberIDs, Collection<Long> inviteIDs, String gbookID) {
+  public Club(String id, String name, String description, String announcement, List<String> posts, 
+      String ownerID, List<String> memberIDs, List<String> inviteIDs, String gbookID) {
         this.id = id;
         this.name = name;
         this.description = description;
@@ -35,15 +29,23 @@ public class Club {
         this.gbookID = gbookID;
   }
 
-  public Club(long id, String name, String description, String announcement, long ownerID, String gbookID){
-    this(id, name, description, announcement, new ArrayList<String>(), ownerID, new HashSet<Long>(), new HashSet<Long>(), gbookID);
+  public Club(String id, String name, String description, String ownerID, String gbookID){
+    this(id, name, description, "", new ArrayList<String>(), ownerID, new ArrayList<String>(), new ArrayList<String>(), gbookID);
   }
 
-  public Club(long id, String name, String description, long ownerID, String gbookID) {
-    this(id, name, description, "", ownerID, gbookID);
+  public Club(String id, String name, String ownerID, String gbookID) {
+    this(id, name, "", ownerID, gbookID);
+  }
+
+  public Club(String id, String name, String ownerID) {
+    this(id, name, "", ownerID, "");
+  }
+
+  public Club() {
+    this("", "", "");
   }
  
-  public long getID() {
+  public String getID() {
     return this.id;
   }
  
@@ -71,7 +73,7 @@ public class Club {
     this.announcement = announcement;
   }
 
-  public Collection<String> getPosts() {
+  public List<String> getPosts() {
     return this.posts;
   }
 
@@ -83,53 +85,53 @@ public class Club {
     this.posts.clear();
   }
 
-  public long getOwnerID() {
+  public String getOwnerID() {
     return ownerID;
   }
 
-  public void setOwnerID(long ownerID) {
+  public void setOwnerID(String ownerID) {
     this.ownerID = ownerID;
   }
  
-  public Collection<Long> getMemberIDs() {
+  public List<String> getMemberIDs() {
     return this.memberIDs;
   }
 
-  public void removeMember(long memberID) {
+  public void removeMember(String memberID) {
     memberIDs.remove(memberID);
   }
 
-  public void removeMembers(Collection<Long> memberIDs) {
-    for (Long l : memberIDs) {
-      this.memberIDs.remove(l);
+  public void removeMembers(List<String> memberIDs) {
+    for (String m : memberIDs) {
+      this.memberIDs.remove(m);
     }
   }
  
-  public Collection<Long> getInviteIDs() {
+  public List<String> getInviteIDs() {
     return this.inviteIDs;
   }
 
-  public void invite(long inviteID) {
+  public void invite(String inviteID) {
     inviteIDs.add(inviteID);
   }
 
-  public void invite(Collection<Long> inviteIDs) {
-    for (Long l : inviteIDs) {
-      this.inviteIDs.add(l);
+  public void invite(List<String> inviteIDs) {
+    for (String i : inviteIDs) {
+      this.inviteIDs.add(i);
     }
   }
 
-  public void uninvite(Long uninviteID) {
+  public void uninvite(String uninviteID) {
     inviteIDs.remove(uninviteID);
   }
 
-  public void uninvite(Collection<Long> uninviteIDs) {
-    for (Long l : uninviteIDs) {
-      this.inviteIDs.remove(l);
+  public void uninvite(List<String> uninviteIDs) {
+    for (String u : uninviteIDs) {
+      this.inviteIDs.remove(u);
     }
   }
 
-  public boolean addMember(long memberID) {
+  public boolean addMember(String memberID) {
     if (inviteIDs.contains(memberID)) {
       inviteIDs.remove(memberID);
       memberIDs.add(memberID);
--- a/capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
@@ -47,37 +47,43 @@ public class ClubServlet extends HttpServlet {
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    List<Club> result = new ArrayList<>();
+    List<Club> clubsRetrieved = new ArrayList<>();
     if (request.getParameter("id") != null) {
       DocumentReference docRef = clubs.document(request.getParameter("id"));
-      ApiFuture<DocumentSnapshot> future = docRef.get();
+      ApiFuture<DocumentSnapshot> asyncDocument = docRef.get();
       DocumentSnapshot document = null;
       Club club = null;
       try {
-        document = future.get();
+        document = asyncDocument.get();
         if (document.exists()) {
           club = document.toObject(Club.class);
         } else {
           System.err.println("Error: no such document!");
+          response.sendError(HttpServletResponse.SC_NOT_FOUND);
+          return;
         }
       } catch (Exception e) {
         System.err.println("Error: " + e);
+        response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+        return;
       }
-      result.add(club);
+      clubsRetrieved.add(club);
     }
     else {
-      ApiFuture<QuerySnapshot> future = clubs.get();
+      ApiFuture<QuerySnapshot> asyncDocument = clubs.get();
       try {
-        List<QueryDocumentSnapshot> documents = future.get().getDocuments();
+        List<QueryDocumentSnapshot> documents = asyncDocument.get().getDocuments();
         for (QueryDocumentSnapshot document : documents) {
-          result.add(document.toObject(Club.class));
+          clubsRetrieved.add(document.toObject(Club.class));
         }
       } catch (Exception e) {
         System.err.println("Error: " + e);
+        response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+        return;
       }
     }
     response.setContentType("application/json;");
-    response.getWriter().println(gson.toJson(result));
+    response.getWriter().println(gson.toJson(clubsRetrieved));
   }
 
   @Override
@@ -97,11 +103,13 @@ public class ClubServlet extends HttpServlet {
     }
 
     Club club = new Club(id, name, description, ownerID, gbookID);
-    ApiFuture<WriteResult> future = clubs.document(id).set(club);
+    ApiFuture<WriteResult> asyncDocument = clubs.document(id).set(club);
     try {
-      System.out.println("Update time : " + future.get().getUpdateTime());
+      System.out.println("Update time : " + asyncDocument.get().getUpdateTime());
     } catch (Exception e) {
       System.err.println("Error: " + e);
+      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+      return;
     }
     response.setContentType("application/json;");
     response.getWriter().println(gson.toJson(club));
@@ -112,15 +120,40 @@ public class ClubServlet extends HttpServlet {
     JsonObject jsonObject = Utility.createRequestBodyJson(request);
     String id = jsonObject.get("id").getAsString();
     Map<String, Object> update = new HashMap<>();
+    DocumentReference docRef = clubs.document(request.getParameter("id"));
+    ApiFuture<DocumentSnapshot> asyncDocument = docRef.get();
+    DocumentSnapshot document = null;
+    Club club = null;
+
+    try {
+      document = asyncDocument.get();
+      if (document.exists()) {
+        club = document.toObject(Club.class);
+      } else {
+        System.err.println("Error: no such document!");
+        response.sendError(HttpServletResponse.SC_NOT_FOUND);
+        return;
+      }
+    } catch (Exception e) {
+      System.err.println("Error: " + e);
+      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+      return;
+    }
 
     if (jsonObject.get("name") != null) {
-      update.put("name", jsonObject.get("name").getAsString());
+      String name = jsonObject.get("name").getAsString();
+      club.setName(name)
+      update.put("name", name);
     }
     if (jsonObject.get("description") != null) {
-      update.put("description", jsonObject.get("description").getAsInt());
+      String description = jsonObject.get("description").getAsString();
+      club.setDescription(description);
+      update.put("description", description);
     }
     if (jsonObject.get("gbookID") != null) {
-      update.put("gbookID", jsonObject.get("gbookID").getAsString());
+      String gbookID = jsonObject.get("gbookID").getAsString();
+      club.setGbookID(gbookID);
+      update.put("gbookID", gbookID);
     }
 
     ApiFuture<WriteResult> writeResult = clubs.document(id).set(update, SetOptions.merge());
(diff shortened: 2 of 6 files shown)


capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -1,166 +1,134 @@
...
-    clubEntity.setProperty("posts" , new ArrayList<String>());
-    clubEntity.setProperty("ownerID", ownerID);
-    clubEntity.setProperty("inviteIDs", inviteIDs);
-    clubEntity.setProperty("memberIDs", new ArrayList<Long>());
-    clubEntity.setProperty("gbookID", gbookID);
-
-    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    datastore.put(clubEntity);
+    Club club = new Club(id, name, description, ownerID, gbookID);
+    ApiFuture<WriteResult> future = clubs.document(id).set(club);
```
antoniolinhart: It looks like if a club with a given `id` already exists in the database it will just be overridden. Do we want this functionality or should we use the `update` like you did in your doPut so we don't lose all the data if we accidentally use the same `id`?

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -1,166 +1,134 @@
...
+    try {
+      System.out.println("Update time : " + future.get().getUpdateTime());
+    } catch (Exception e) {
+      System.err.println("Error: " + e);
+    }
+    Gson gson = new Gson();
```
antoniolinhart: I think you can remove this and use your field `gson` object

capstone-backend/src/main/java/com/google/sps/data/Club.java
```diff
@@ -35,15 +29,23 @@ public Club(long id, String name, String description, String announcement, Colle
...
+    this(id, name, "", ownerID, gbookID);
+  }
+
+  public Club(String id, String name, String ownerID) {
+    this(id, name, "", ownerID, "");
   }
 
-  public Club(long id, String name, String description, long ownerID, String gbookID) {
-    this(id, name, description, "", ownerID, gbookID);
+  public Club() {
```
antoniolinhart, line 44: I know yesterday we talked about this a bit offline - but I was curious what the use of the default constructor is for Firestore?

antoniolinhart (approved):
Looks good! Nice job refactoring the tests and allowing us to add clubs to Firestore

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: You're right, forgot to take this out. Thanks.

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: I think for now we can leave it alone, since eventually we won't allow users to enter their own IDs and rather generate them ourselves (likely via some sort of Hash function).

capstone-backend/src/main/java/com/google/sps/data/Club.java
steven-solar: I believe it lets Firestore cast the objects it retrieves (cast to type Club in this case). I believe you can just use the default Java constructor too.

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -1,166 +1,134 @@
...
-      Collection<Long> inviteIDs = (HashSet<Long>) e.getProperty("inviteIDs");
-
-      long ownerID = (long) e.getProperty("ownerID");
-      String gbookID = (String) e.getProperty("gbookID");
-      Club club = new Club(id, name, description, announcement, ownerID, gbookID);
-      clubs.add(club);
+    List<Club> result = new ArrayList<>();
+    if (request.getParameter("id") != null) {
+      DocumentReference docRef = clubs.document(request.getParameter("id"));
+      ApiFuture<DocumentSnapshot> future = docRef.get();
```
geoffbass: nit: Replace `future` with a more specific name

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -1,166 +1,134 @@
...
-      Collection<Long> memberIDs = (HashSet<Long>) e.getProperty("memberIDs");
-      
-      @SuppressWarnings("unchecked") // Cast can't verify generic type.
-      Collection<Long> inviteIDs = (HashSet<Long>) e.getProperty("inviteIDs");
-
-      long ownerID = (long) e.getProperty("ownerID");
-      String gbookID = (String) e.getProperty("gbookID");
-      Club club = new Club(id, name, description, announcement, ownerID, gbookID);
-      clubs.add(club);
+    List<Club> result = new ArrayList<>();
```
geoffbass: nit: Replace `result` with a more specific name

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -1,166 +1,134 @@
...
+      Club club = null;
+      try {
+        document = future.get();
+        if (document.exists()) {
+          club = document.toObject(Club.class);
+        } else {
+          System.err.println("Error: no such document!");
+        }
+      } catch (Exception e) {
+        System.err.println("Error: " + e);
```
geoffbass: It looks like you're logging the errors here but then still sending a success response with an empty payload. It would probably be better to respond with specific error codes (404, 500, etc.).

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -1,166 +1,134 @@
...
+      }
+      result.add(club);
+    }
+    else {
+      ApiFuture<QuerySnapshot> future = clubs.get();
```
geoffbass: nit: Rename `future`

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -1,166 +1,134 @@
...
+      try {
+        List<QueryDocumentSnapshot> documents = future.get().getDocuments();
+        for (QueryDocumentSnapshot document : documents) {
+          result.add(document.toObject(Club.class));
+        }
+      } catch (Exception e) {
+        System.err.println("Error: " + e);
```
geoffbass, line 80: Same here: if there is an exception you probably want to surface some info about it to the client.

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
geoffbass: nit: Rename `future`

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -1,166 +1,134 @@
...
-    clubEntity.setProperty("gbookID", gbookID);
-
-    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    datastore.put(clubEntity);
+    Club club = new Club(id, name, description, ownerID, gbookID);
+    ApiFuture<WriteResult> future = clubs.document(id).set(club);
+    try {
+      System.out.println("Update time : " + future.get().getUpdateTime());
+    } catch (Exception e) {
+      System.err.println("Error: " + e);
```
geoffbass, line 110: Add some logic for the case of I/O failure here. You could optionally retry the request and ultimately respond with an error code if needed.

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -1,166 +1,134 @@
...
-    clubEntity.setProperty("ownerID", ownerID);
-    clubEntity.setProperty("gbookID", gbookID);
-
-    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    datastore.put(clubEntity);
+    ApiFuture<WriteResult> writeResult = clubs.document(id).set(update, SetOptions.merge());
+    try {
+      System.out.println("Update time : " + writeResult.get().getUpdateTime());
+    } catch (Exception e) {
+      System.err.println("Error: " + e);
```
geoffbass, line 163: Same here: probably need some additional error-handling logic. You might also need to send the updated document in the response payload to handle updating the front-end without a refresh, though you could also do optimistic front-end updates and notify the user when changes aren't saved.

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -1,10 +1,27 @@
 import React, { Component } from 'react';
 
 class ClubPage extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      response: {}
```
geoffbass: nit: This should probably be `club`. Think about the name from the template's point of view, i.e. what data is being rendered. Glad to see you initializing state properties though!

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -1,10 +1,27 @@
...
+    }
+  }
+
+  fetchClubData = () => {
+    fetch(`/api/clubs?id=${this.props.match.params.id}`)
+        .then(response => response.json()).then(res => this.setState({response: res[0]}));
+  }
+
   render() {
+    this.fetchClubData();
```
geoffbass, line 18: This should probably be in `componentDidMount` (and also probably `componentDidUpdate` in case `match.params.id` changes). Those are generally the two places you want to make API calls. Keep logic and side-effects out of your `render` method whenever possible. It also looks to me like there might be an infinite `render` -> `fetchClubData` -> `setState` -> `render`... loop here: you can verify that by looking at the Network tab in the Dev Tools. If so, using `componentDidMount` should fix that since it only gets invoked once.

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -1,10 +1,27 @@
...
     return (
       <div>
-        <div> {this.props.match.params.id} </div>
+        <div> Club Name: {this.state.response.name} </div>
+        <div> ID: {this.props.match.params.id} </div>
+        <div> Description: {this.state.response.description} </div>
+        <div> OwnerID: {this.state.response.ownerID} </div>
+        <div> GbookID: {this.state.response.gbookID} </div>
```
geoffbass: Not necessary to do right now since this is basically just dumping the data from the API onto the DOM, but don't forget to add tests for this component once you flesh it out!

geoffbass (changes requested):

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: I wasn't sure what to name it since an ApiFuture represents the result of an asynchronous computation (which is resolved when I call future.get()), so it isn't really anything but an intermediary. I'll rename it something like ``` asyncDocument ``` so that it is more informative for the reader.

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: I'll rename this ``` clubsRetrieved ``` instead.

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: Good idea, I'll send a 404 for when ``` !document.exists() ``` and a 500 for my generic Exception catch. Seems the servlets defined final static variables for different codes, (https://www.javamex.com/tutorials/servlets/http_status_code.shtml) which I'll use. I'll also return immediately to kick out of the function in this case.

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: Will do!

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: I'll send a Http.SC_INTERNAL_SERVER_ERROR (500) in this case.

steven-solar: * meant to say HttpServletResponse.SC_INTERNAL_SERVER_ERROR

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: That's a good idea. I'm not sure how to trigger a "retry" but I can look into it for future iterations, for now I'll just send a 500. I just want to get the base functionality down, as I'll be abstracting much of this in the next PR.

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: That's a good idea! I'll add that.

steven-solar: I did that in my Post, but forgot to include in my Put, will add.

capstone-frontend/src/components/ClubPage.jsx
steven-solar: Will change, and thanks!

capstone-frontend/src/components/ClubPage.jsx
steven-solar: Good idea, I'll create both those functions and have them call ``` fetchClubData() ```.

capstone-frontend/src/components/ClubPage.jsx
steven-solar: Will do! I didn't bother adding testing here since I knew this was just a temporary display and would change a ton, but will definitely incorporate in the future.

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -1,166 +1,171 @@
...
+      ApiFuture<DocumentSnapshot> asyncDocument = docRef.get();
+      DocumentSnapshot document = null;
+      Club club = null;
+      try {
+        document = asyncDocument.get();
+        if (document.exists()) {
+          club = document.toObject(Club.class);
+        } else {
+          System.err.println("Error: no such document!");
+          response.sendError(HttpServletResponse.SC_NOT_FOUND);
```
geoffbass, line 62: Nice!

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -1,166 +1,171 @@
...
+          return;
+        }
+      } catch (Exception e) {
+        System.err.println("Error: " + e);
+        response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
```
geoffbass, line 67: Looks good

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -1,10 +1,34 @@
...
+    fetch(`/api/clubs?id=${this.props.match.params.id}`)
+        .then(response => response.json()).then(res => this.setState({club: res[0]}));
+  }
+
+  componentDidMount() {
+    this.fetchClubData();
+  }
+
+  componentDidUpdate() {
+    this.fetchClubData();
```
geoffbass: I think you'll want to check that the relevant prop changed before calling this, otherwise you'll end up in the same kind of render loop. [`componentDidUpdate`](https://reactjs.org/docs/react-component.html#componentdidupdate) gives you access to `prevProps` for this purpose.

geoffbass (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: Thanks!

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: Thanks!

capstone-frontend/src/components/ClubPage.jsx
steven-solar: I didn't realize that, but that makes a lot of sense. Changing this to check for a change in state before calling fetchData.

capstone-backend/src/main/java/com/google/sps/data/Club.java
```diff
@@ -1,29 +1,23 @@
...
+  private final String id;
   private String name;
   private String description;
   private String announcement;
-  private Collection<String> posts;
-  private long ownerID;
-  private Collection<Long> memberIDs;
-  private Collection<Long> inviteIDs;
+  private List<String> posts;
+  private String ownerID;
```
MauroGuerrero20, line 14: Will there only be one owner to a Club? If not maybe make this a List

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -1,166 +1,171 @@
...
+
 import java.io.BufferedReader;
 import java.io.InputStreamReader;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.HashMap;
```
MauroGuerrero20, line 28:
```import com.google.gson.GsonBuilder;
import com.google.gson.JsonParser;
import java.util.Collection;
```

Unnecessary imports

capstone-backend/src/test/java/com/google/sps/data/ClubTest.java
```diff
@@ -6,6 +6,7 @@
 import java.util.Arrays;
 import java.util.List;
 
+
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.BeforeClass;
```
MauroGuerrero20, line 12: Remove imports 
```
import org.junit.BeforeClass;
import com.google.sps.data.Club;

```

capstone-frontend/package.json
```diff
@@ -6,6 +6,7 @@
     "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/react": "^9.3.2",
     "@testing-library/user-event": "^7.1.2",
+    "deep-equal": "^2.0.3",
```
MauroGuerrero20, line 9: Is there a reason why we need the "deep-equal" dependency? It seems that you don't use it on the React frontend

MauroGuerrero20 (changes requested):
A few minor changes required.

capstone-backend/src/main/java/com/google/sps/data/Club.java
steven-solar: That was the plan for now, but we can change it in the future if we think that's best. Maybe there are "admins" of Clubs (thinking of a Clash of Clans set up). Will leave for now.

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: Good catch! Removed.

capstone-backend/src/test/java/com/google/sps/data/ClubTest.java
steven-solar: Removed.

capstone-frontend/package.json
steven-solar: I was testing out the package for use in checking equivalency between JSONs, and then forgot to remove it when I decided not to use it. Will remove.

MauroGuerrero20 (approved):
LGTM

PR #37, opened 2020-07-02
Minor change to max results calculation

antoniolinhart:
Realized I had not fully responded to one of @VivekShankar's comments on my last PR where I was calling parseNaturalNumber twice when it was not necessary.

### DIFF
--- a/capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
@@ -50,10 +50,10 @@ public class SearchServlet extends HttpServlet {
     searchTerm = URLEncoder.encode(searchTerm, "UTF-8");
 
     int maxResults = 0;
-    if (request.getParameter("maxResults") != null) { 
-      maxResults = parseNaturalNumber(request.getParameter("maxResults"));
-    } 
-    if (maxResults == -1) {
+    String maxResultsParam = request.getParameter("maxResults"); 
+    if (maxResultsParam != null) { 
+      maxResults = parseNaturalNumber(maxResultsParam);
+    } else if (maxResultsParam == null || maxResults == -1) {
       maxResults = DEFAULT_MAX_RESULTS;
     }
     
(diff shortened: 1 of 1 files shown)


capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -49,11 +49,11 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-    int maxResults = parseNaturalNumber(request.getParameter("maxResults"));
-    if (request.getParameter("maxResults") != null) { 
-      maxResults = parseNaturalNumber(request.getParameter("maxResults"));
-    } 
-    if (maxResults == -1) {
+    int maxResults = 0;
+    String maxResultsParam = request.getParameter("maxResults"); 
+    if (maxResultsParam != null) { 
+      maxResults = parseNaturalNumber(maxResultsParam);
+    } else if (maxResultsParam == null || maxResults == -1) {
```
steven-solar: What happens if `maxResultsParam = `"-1" or some other invalid string (could contain letters, be a negative number, be the empty string, etc)? It would satisfy the first `if` clause (as it is not `null`), but then it would not hit the `else if`, and thus `maxResults` would be set to `-1` instead of `DEFAULT_MAX_RESULTS`. Changing the `else if` to just an `if` clause would fix this problem.

steven-solar (changes requested):

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
antoniolinhart: Yup! This makes sense. I thought there would be a cleaner way to do it where I'm not doing specific null/not null checks by using an else block but the way I wrote it is not what I intended.

Thank you!

steven-solar (approved):
Looks great!

MauroGuerrero20 (approved):
LGTM

VivekShankar (approved):
LGTM

PR #38, opened 2020-07-03
Generalized get method

### LINKED ISSUE #35: Generic GET/POST/PUT Database Access

**Additional context**

steven-solar:
This PR creates a generalized `get()` method in the Utility class for all servlets to use in the future. 
* It allows the requester to apply any number of queries both for equality in primitive type and string classes, and membership in object fields of type `List` or `ArrayList`. Query parameters must match object field names exactly, or errors will be thrown and a null `List` will be returned. Reflection and the `Field` class and methods are used to achieve this. 
* If no queries are applied, the entire collection of documents is returned. 
* If a valid ID is supplied, a singleton list with the matching object is returned, if no object is found errors will be thrown and a null `List` will be returned. 

It will cast the retrieved documents to be objects of the given Class. `GenericClass` and the built in Java generic type are used for this purpose, allowing retrieval of object type at runtime. 

All servlets can call the generalized `get()` method to perform queries for equality or membership on any field of any object class. The method performs all necessary error handling (on errors, bad requests, or when an id is not found), the method will attach response headers of 400, 404 and 500 appropriately, and return a null List<T>). Servlets should call this method, and print its results to the HttpServletResponse. 

## Example Use
* `/api/clubs?name=example%20name` will filter for all clubs with the name "example name" 
* `/api/clubs?memberIDs=member1` will filter for all clubs where "member1" is a memberID of a club member
* `/api/clubs/?name=example%20name&memberIDs=member1` will filter for all clubs with the name "example name" and where "member1" is a memberID of a club member
* `/api/clubs?id=exampleID` will return a singleton list with the club whose id is "exampleID"
* `/api/clubs` will return all clubs

### DIFF
--- a/capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
@@ -47,35 +47,7 @@ public class ClubServlet extends HttpServlet {
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    List<Club> result = new ArrayList<>();
-    if (request.getParameter("id") != null) {
-      DocumentReference docRef = clubs.document(request.getParameter("id"));
-      ApiFuture<DocumentSnapshot> future = docRef.get();
-      DocumentSnapshot document = null;
-      Club club = null;
-      try {
-        document = future.get();
-        if (document.exists()) {
-          club = document.toObject(Club.class);
-        } else {
-          System.err.println("Error: no such document!");
-        }
-      } catch (Exception e) {
-        System.err.println("Error: " + e);
-      }
-      result.add(club);
-    }
-    else {
-      ApiFuture<QuerySnapshot> future = clubs.get();
-      try {
-        List<QueryDocumentSnapshot> documents = future.get().getDocuments();
-        for (QueryDocumentSnapshot document : documents) {
-          result.add(document.toObject(Club.class));
-        }
-      } catch (Exception e) {
-        System.err.println("Error: " + e);
-      }
-    }
+    List<Club> result = Utility.get(clubs, request, new GenericClass(Club.class));
     response.setContentType("application/json;");
     response.getWriter().println(gson.toJson(result));
   }
--- a/capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
@@ -43,41 +43,7 @@ public class ClubServlet extends HttpServlet {
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    List<Club> clubsRetrieved = new ArrayList<>();
-    if (request.getParameter("id") != null) {
-      DocumentReference docRef = clubs.document(request.getParameter("id"));
-      ApiFuture<DocumentSnapshot> asyncDocument = docRef.get();
-      DocumentSnapshot document = null;
-      Club club = null;
-      try {
-        document = asyncDocument.get();
-        if (document.exists()) {
-          club = document.toObject(Club.class);
-        } else {
-          System.err.println("Error: no such document!");
-          response.sendError(HttpServletResponse.SC_NOT_FOUND);
-          return;
-        }
-      } catch (Exception e) {
-        System.err.println("Error: " + e);
-        response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
-        return;
-      }
-      clubsRetrieved.add(club);
-    }
-    else {
-      ApiFuture<QuerySnapshot> asyncDocument = clubs.get();
-      try {
-        List<QueryDocumentSnapshot> documents = asyncDocument.get().getDocuments();
-        for (QueryDocumentSnapshot document : documents) {
-          clubsRetrieved.add(document.toObject(Club.class));
-        }
-      } catch (Exception e) {
-        System.err.println("Error: " + e);
-        response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
-        return;
-      }
-    }
+    List<Club> result = Utility.get(clubs, request, new GenericClass(Club.class));
     response.setContentType("application/json;");
     response.getWriter().println(gson.toJson(clubsRetrieved));
   }
--- a/capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
@@ -43,9 +43,11 @@ public class ClubServlet extends HttpServlet {
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    List<Club> result = Utility.get(clubs, request, new GenericClass(Club.class));
-    response.setContentType("application/json;");
-    response.getWriter().println(gson.toJson(clubsRetrieved));
+    List<Club> retrievedClubs = Utility.get(clubs, request, response, new GenericClass(Club.class));
+    if (retrievedClubs != null) {
+      response.setContentType("application/json;");
+      response.getWriter().println(gson.toJson(retrievedClubs));
+    }
   }
 
   @Override
--- a/capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
@@ -66,7 +66,7 @@ public class ClubServlet extends HttpServlet {
       gbookID = jsonObject.get("gbookID").getAsString();
     }
 
-    Club club = new Club(id, name, description, ownerID, gbookID);
+    Club club = new Club(name, description, ownerID, gbookID);
     ApiFuture<WriteResult> asyncDocument = clubs.document(id).set(club);
     try {
       System.out.println("Update time : " + asyncDocument.get().getUpdateTime());
--- a/capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
@@ -47,9 +47,11 @@ public class ClubServlet extends HttpServlet {
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    List<Club> result = Utility.get(clubs, request, new GenericClass(Club.class));
-    response.setContentType("application/json;");
-    response.getWriter().println(gson.toJson(result));
+    List<Club> retrievedClubs = Utility.get(clubs, request, response, new GenericClass(Club.class));
+    if (retrievedClubs != null) {
+      response.setContentType("application/json;");
+      response.getWriter().println(gson.toJson(retrievedClubs));
+    }
   }
 
   @Override
--- a/capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
@@ -70,7 +70,7 @@ public class ClubServlet extends HttpServlet {
       gbookID = jsonObject.get("gbookID").getAsString();
     }
 
-    Club club = new Club(id, name, description, ownerID, gbookID);
+    Club club = new Club(name, description, ownerID, gbookID);
     ApiFuture<WriteResult> future = clubs.document(id).set(club);
     try {
       System.out.println("Update time : " + future.get().getUpdateTime());
--- a/capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
@@ -14,15 +14,11 @@ import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
-import com.google.gson.JsonParser;
 
 import com.google.sps.data.Club;
 
-import java.io.BufferedReader;
-import java.io.InputStreamReader;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.HashMap;
--- a/capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
@@ -66,6 +66,14 @@ public class Utility {
     return jsonObject;
   }
 
+  /**
+   * Query the given collection by id. 
+   * @param {collectionReference} reference to the appropriate database collection
+   * @param {request} request sent to the backend
+   * @param {response} response returned from the call
+   * @param {genericClass} a Generic class, used in casting documents from the database
+   * @return List<T> a singleton List of the Object that matches the ID in the request.
+   */
   private static <T> List<T> getById(CollectionReference collectionReference, HttpServletRequest request, 
       HttpServletResponse response, GenericClass<T> genericClass) throws IOException {
     if (request.getParameterMap().size() > 1) {
@@ -97,6 +105,14 @@ public class Utility {
     return retrievedObjects;
   }
 
+  /**
+   * Appends filters to the given query depending on fields in the request query 
+   * @param {genericClass} a Generic class, used in casting documents from the database
+   * @param {response} response returned from the call
+   * @param {it} iterator for the Map of query parameter names to values
+   * @param {query} query being modified, to eventually be used to query the database.
+   * @return Query the query being modified, to eventually be used to query the database.
+   */
   private static <T> Query addToQuery(GenericClass<T> genericClass, HttpServletResponse response, 
       Iterator it, Query query) throws IOException {
     Map.Entry<String, String[]> entry = (Map.Entry) it.next();
@@ -136,6 +152,14 @@ public class Utility {
     }
   }
 
+  /**
+   * Gets an asynchronous collection of matching documents from the database matching the request
+   * @param {collectionRefence} reference to the appropriate collection of documents in the database
+   * @param {genericClass} a Generic class, used in casting documents from the database
+   * @param {request} request sent to the backend
+   * @param {response} response returned from the call
+   * @return Query the query being modified, to eventually be used to query the database.
+   */
   private static <T> ApiFuture<QuerySnapshot> getByField(CollectionReference collectionReference, HttpServletRequest request, 
       HttpServletResponse response, GenericClass<T> genericClass) throws IOException {
     Iterator it = request.getParameterMap().entrySet().iterator();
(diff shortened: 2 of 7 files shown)


capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -21,8 +22,14 @@
 import java.io.BufferedReader;
 import java.io.InputStreamReader;
 import java.io.IOException;
+import java.lang.reflect.Field;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.Map;
```
antoniolinhart: nit: these 4 imports aren't in alphabetical order, it makes the diffs more consistent

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -58,4 +65,181 @@ public static JsonObject createRequestBodyJson(HttpServletRequest request) {
...
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+      return null;
+    }
+    if (entry.getValue()[0].isEmpty()) {
+      System.err.println("Error: The parameter value cannot be empty");
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+      return null;
+    }
+    String parameterValue = entry.getValue()[0];
+    Field[] fields = genericClass.getMyType().getDeclaredFields();
```
antoniolinhart, line 170: Good job considering the possible failures in this section!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -58,4 +65,181 @@ public static JsonObject createRequestBodyJson(HttpServletRequest request) {
...
+      return getById(collectionReference, request, response, genericClass);
+    }
+    if (request.getParameterMap().size() == 0) {
+      asyncQuery = collectionReference.get();
+    }
+    else {
+      asyncQuery = getByField(collectionReference, request, response, genericClass);
+      if (asyncQuery == null) {
+        return null;
+      }
```
antoniolinhart, line 222: I like how you separated these 3 cases into different methods for when you're just given an id, when you're given no parameters and when you are given a specific list of fields!

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -47,37 +47,11 @@ public ClubServlet() throws IOException {
...
-          System.err.println("Error: no such document!");
-        }
-      } catch (Exception e) {
-        System.err.println("Error: " + e);
-      }
-      result.add(club);
+    List<Club> retrievedClubs = Utility.get(clubs, request, response, new GenericClass(Club.class));
+    if (retrievedClubs != null) {
+      response.setContentType("application/json;");
+      response.getWriter().println(gson.toJson(retrievedClubs));
```
antoniolinhart, line 49: Much cleaner get method!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -58,4 +65,181 @@ public static JsonObject createRequestBodyJson(HttpServletRequest request) {
...
+    if (entry.getValue().length != 1) {
+      System.err.println("Error: Each parameter should have only one value");
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+      return null;
+    }
+    if (entry.getValue()[0].isEmpty()) {
+      System.err.println("Error: The parameter value cannot be empty");
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+      return null;
+    }
```
antoniolinhart, line 115: In future PRs we should consider how we want to display these error codes to users on the frontend

antoniolinhart (approved):
LGTM!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Oops, grouped List and ArrayList together in my head and forgot to alphabetize them. Will fix!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Thanks!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Thanks! These seemed to be the three major cases I could think of, not sure if there's any other scenarios to be considered.

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: Thank you! Should be super easy for all our servlets to use this in the future.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Good idea. For now I just pop up an alert saying "Looks like we're having trouble connecting to our database, hang tight" on failed posts, but definitely need to build that out in the future

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -58,4 +65,181 @@ public static JsonObject createRequestBodyJson(HttpServletRequest request) {
     }
     return jsonObject;
   }
+
```
MauroGuerrero20, line 54: The static method ```createRequestBodyJson``` (Line 46-67) can be simplified to two lines. It also contains deprecated code. 


#### For example
```
String googleUserTokenObjString = request.getReader().lines().collect(Collectors.joining());
JsonObject googleUserTokenObjJSON = JsonParser.parseString(googleUserTokenObjString).getAsJsonObject();
```

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -58,4 +65,181 @@ public static JsonObject createRequestBodyJson(HttpServletRequest request) {
...
+  private static <T> List<T> getById(CollectionReference collectionReference, HttpServletRequest request, 
+      HttpServletResponse response, GenericClass<T> genericClass) throws IOException {
+    if (request.getParameterMap().size() > 1) {
+      System.err.println("Error: No other parameter can be sent with an ID");
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+      return null;
+    }
+    List<T> retrievedObjects = new ArrayList<>();
+    ApiFuture<QuerySnapshot> asyncQuery;
+    DocumentReference docRef = collectionReference.document(request.getParameter("id"));
```
MauroGuerrero20, line 71: ```asyncQuery``` is never used

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -58,4 +65,181 @@ public static JsonObject createRequestBodyJson(HttpServletRequest request) {
...
+   * @param {response} response returned from the call
+   * @param {it} iterator for the Map of query parameter names to values
+   * @param {query} query being modified, to eventually be used to query the database.
+   * @return Query the query being modified, to eventually be used to query the database.
+   */
+  private static <T> Query addToQuery(GenericClass<T> genericClass, HttpServletResponse response, 
+      Iterator it, Query query) throws IOException {
+    Map.Entry<String, String[]> entry = (Map.Entry) it.next();
+    String parameterName = entry.getKey();
+    if (entry.getValue().length != 1) {
```
MauroGuerrero20, line 106: Gitpod gives us a me a warning about ```(Map.Entry) it.next()``` 

Type safety: The expression of type Map.Entry needs unchecked conversion to conform to Map.Entry<String,String[]>

It also gives me a warning about ```Iterator```

Iterator is a raw type. References to generic type Iterator<E> should be parameterized

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -58,4 +65,181 @@ public static JsonObject createRequestBodyJson(HttpServletRequest request) {
...
+   * @param {request} request sent to the backend
+   * @param {response} response returned from the call
+   * @return Query the query being modified, to eventually be used to query the database.
+   */
+  private static <T> ApiFuture<QuerySnapshot> getByField(CollectionReference collectionReference, HttpServletRequest request, 
+      HttpServletResponse response, GenericClass<T> genericClass) throws IOException {
+    Iterator it = request.getParameterMap().entrySet().iterator();
+    Map.Entry<String, String[]> entry = (Map.Entry) it.next();
+    String parameterName = entry.getKey();
+    if (entry.getValue().length != 1) {
```
MauroGuerrero20, line 159: Same thing here

MauroGuerrero20 (commented):
Minor Optimization Changes

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Good catch, forgot to remove that. Thanks!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Cool! Forgot to change the JsonParser to utilize the newer static methods, but this is a great solution!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: I'll parameterize the iterator, good catch! Now that I've parameterized the iterator, I can remove the forced cast to type Map.Entry and that should deal with the Type Safety warning. Thanks for the comment!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Fixed!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -36,26 +44,185 @@ public static Firestore getFirestoreDb() throws IOException {
...
+    }
+  }
+
+  /**
+   * Gets an asynchronous collection of matching documents from the database matching the request
+   * @param {collectionRefence} reference to the appropriate collection of documents in the database
+   * @param {genericClass} a Generic class, used in casting documents from the database
+   * @param {request} request sent to the backend
+   * @param {response} response returned from the call
+   * @return Query the query being modified, to eventually be used to query the database.
```
VivekShankar: Can you update the `@return` annotation here and reorder the `@param` annotations to match the parameters in the `getByField` method?

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -36,26 +44,185 @@ public static Firestore getFirestoreDb() throws IOException {
...
+   */
+  private static <T> ApiFuture<QuerySnapshot> getByField(CollectionReference collectionReference, HttpServletRequest request, 
+      HttpServletResponse response, GenericClass<T> genericClass) throws IOException {
+    Iterator<Map.Entry<String, String[]>> it = request.getParameterMap().entrySet().iterator();
```
VivekShankar, line 151: Was there a specific reason that you wanted to use an `Iterator` here? How about using a for loop to loop over  the keys in the map?

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -36,26 +44,185 @@ public static Firestore getFirestoreDb() throws IOException {
...
+
+  /**
+   * Appends filters to the given query depending on fields in the request query 
+   * @param {genericClass} a Generic class, used in casting documents from the database
+   * @param {response} response returned from the call
+   * @param {it} iterator for the Map of query parameter names to values
+   * @param {query} query being modified, to eventually be used to query the database.
+   * @return Query the query being modified, to eventually be used to query the database.
+   */
+  private static <T> Query addToQuery(GenericClass<T> genericClass, HttpServletResponse response,
```
VivekShankar: It looks like the logic in the `addToQuery` method here is quite similar to what's already in `getByField`. Is there an easy way to remove the duplicate code (e.g. I'm wondering if a for loop over the `request.getParameterMap()`as opposed to using an `Iterator` would help)?

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -36,26 +44,185 @@ public static Firestore getFirestoreDb() throws IOException {
...
+   * @param {collectionRefence} reference to the appropriate collection of documents in the database
+   * @param {genericClass} a Generic class, used in casting documents from the database
+   * @param {request} request sent to the backend
+   * @param {response} response returned from the call
+   * @return Query the query being modified, to eventually be used to query the database.
+   */
+  private static <T> ApiFuture<QuerySnapshot> getByField(CollectionReference collectionReference, HttpServletRequest request, 
+      HttpServletResponse response, GenericClass<T> genericClass) throws IOException {
+    Iterator<Map.Entry<String, String[]>> it = request.getParameterMap().entrySet().iterator();
+    Map.Entry<String, String[]> entry = it.next();
```
VivekShankar, line 157: I think the iterator is guaranteed to be non-empty here, but it might be helpful to add a quick `hasNext()` check (calling `next` on an empty iterator will throw a `NoSucheElementException) (https://docs.oracle.com/javase/8/docs/api/java/util/Iterator.html#next--).

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -36,26 +44,185 @@ public static Firestore getFirestoreDb() throws IOException {
...
+      if (asyncQuery == null) {
+        return null;
+      }
     }
     try {
-      Gson gson = new Gson();
-      JsonParser parser = new JsonParser();
-      JsonElement data = parser.parse(jb.toString());
-      jsonObject = data.getAsJsonObject();
+      List<QueryDocumentSnapshot> documents = asyncQuery.get().getDocuments();
```
VivekShankar: Not required, but I'd recommend checking out the Java [Streams API](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html) if you haven't used it before. You can use Streams `map` and `collect` methods to do this entire operation in one line.

VivekShankar (approved):
This is cool - love that you're taking advantage of more advanced features of Java such as Generics and Reflection to develop a more general `get` method. Just left a few suggestions; feel free to merge after taking a look!

MauroGuerrero20 (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: I used an `Iterator` because the creation of the `Query` on the first `Entry` is different from adding to the query with every other entry. There didn't seem to be any way to standardize it enough to use one succinct `for` loop, as there is no way to instantiate an "empty query." 

My solution was to at the top create the Query with the first Entry, and then iterate through the rest of the Entries, adding to the Query as the Iterator moves along.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Similar to my answer above, creating the `Query` and adding to it are different, and there is no way to create a base, empty Query to build off of. This forced me to duplicate some code, but I couldn't come up with a nice way around it without running into issues. 

For example, if I used a `for` loop and gave it different behavior on the first loop (instantiate Query on `i == 0`) than on the rest (add to Query on `i != 0`) I received errors along the lines of "Query may not have been instantiated" because there is no way to initialize the Query before the loop. 

I do think that the code here appears messy, and I'm open to discussing alternative ways of writing this.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Thanks, will add!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: I'll check it out! Would be nice to clean it up.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -36,26 +44,185 @@ public static Firestore getFirestoreDb() throws IOException {
...
+    ApiFuture<DocumentSnapshot> asyncDocument = docRef.get();
+    DocumentSnapshot document = null;
+    T item = null;
     try {
-      BufferedReader reader = new BufferedReader(new InputStreamReader((request.getInputStream())));
-      while ((line = reader.readLine()) != null) {
-       jb.append(line);
+      document = asyncDocument.get();
+      if (document.exists()) {
+        item = document.toObject(genericClass.getMyType());
```
antoniolinhart, line 78: Hey Steven -- I found out a solution to our id problem. We should be able to do document.getId() here and get the string id of the given data stored in firebase (https://firebase.google.com/docs/reference/android/com/google/firebase/firestore/DocumentSnapshot). If we can ensure all of our objects in the java backend have setId methods, we could do item.setId() with the new id from the DocumentSnapshot object too!

This also applies to your other get sub-methods. It seems important to fit into this PR instead of waiting, since ids are important and we do need to access them on the frontend

antoniolinhart (commented):
If this id change works, this will be a great way to store and dynamically create ids using firestore's methods!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Sure!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: The problem is that since these methods are written for `Generic` types, there is no way to set the ID of the casted objects before returning them to the servlet. We would have to return a `List<QueryDocumentSnapshot>`, and then cast and set IDs in the servlet, before appending them to the response. This would introduce repeated code in the servlets, as they'd all need to perform this same functionality. 

After our discussion today, I think it would be best to use the Java UUID library instead, and have the primary index in firebase be the object's id set by the UUID functions. This will have some repeated data, as the object id and primary key will be the same, but the simplification of code outweighs this drawback in my mind.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
VivekShankar: Oh I see, thanks for the clarification! Yeah, given that creating the initial `Query` is different from adding to the query on every subsequent iteration and moreover, that you can't initialize an empty query, this approach sounds good to me.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
VivekShankar: Thanks again for the clarification - yeah, given the reasons you've mentioned above I think this looks good!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: After further discussion, I think it's best to use Firebase's IDs. I created a class called BaseEntity that has the needed setId() function, and the generic functions apply to type <T extends BaseEntity> allowing us to set the ID of the generic objects retrieved before they are returned to the servlet. This way, the id is not stored as a field of the object in firebase, but only as the primary key. When it is returned to the frontend, it will have that primary key as its id.

PR #41, opened 2020-07-06
Abstract put

steven-solar:
This PR creates a generalized `put` method in the Utility class for all servlets to use in the future. It allows the requester to supply any number of fields in the body of the `put` request to update in the database. 
* An `id` must be supplied or errors will be thrown. 
* Keys in the body must match object field names exactly, or errors will be thrown and a null Object will be returned. The Field class and methods are used to achieve this. 
* The exception to this rule is for `List` types. In order to add to Lists of type String the put body should contain a field called `add_<list_field_name>`. For example, to add `"member1"` to `memberIDs`, the put body should contain `"add_memberIDs": "member1"`. To remove from Lists of type String the put body should contain a field called `remove_<list_field_name>`. To remove `"member1"` from `memberIDs`, the put body should contain `"remove_memberIDs": "member1"`. As of now, this only supports Lists of type String, since you cannot retrieve the type of a List via reflection, and all our Lists are of type String thus far.

The function will return the updated object in the response body. In order to do this, the `getById` helper function is updated to accept an `id` directly, so the `put` method can call it to retrieve the updated object.

All servlets can call the generalized `put` method to update any field of any object class. The method performs all necessary error handling (on errors, bad requests, or when an id is not found), the method will attach response headers of 400, 404 and 500 appropriately, and return a null object). Servlets should call this method, and print its results to the `HttpServletResponse response`.


capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -236,4 +244,76 @@ public static JsonObject createRequestBodyJson(HttpServletRequest request) throw
...
+    }
+    if (id.length() == 0) {
+      System.err.println("Error caused by either an empty or non-existent \"id\" field in the post body.");
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+      return null;
+    }
+    Map<String, Object> update = new HashMap<>();
+    Field[] fields = genericClass.getMyType().getDeclaredFields();
+    boolean containsParameter = false;
+    for (String key : jsonObject.keySet()) {
```
geoffbass, line 297: Rather than looping over the full set of keys in the request, could you look directly for each field on the JSON object as you iterate through the below loop? That should have the same end result but would avoid looping over fields that don't match the type being updated.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -236,4 +244,76 @@ public static JsonObject createRequestBodyJson(HttpServletRequest request) throw
...
+          }
+        }
+        else if (("remove_"+field.getName()).equals(key)) {
+          if (type.equals("java.util.List") || type.equals("java.util.ArrayList")) {
+            containsParameter = true;
+            update.put(field.getName(), FieldValue.arrayRemove(jsonObject.get(key).getAsString()));
+          }
+        }
+      }
+      if (!containsParameter) {
```
geoffbass, line 334: It looks to me like you could just make this an `else` case and avoid the use of `containsParameter` altogether.

geoffbass (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -81,55 +81,10 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-    try {
-      System.out.println("Update time : " + writeResult.get().getUpdateTime());
-    } catch (Exception e) {
-      System.err.println("Error: " + e);
-      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
-      return;
+    Club updatedClub = (Club) Utility.put(clubs, request, response, new GenericClass(Club.class));
+    if (updatedClub != null) {
+      response.setContentType("application/json;");
+      response.getWriter().println(gson.toJson(updatedClub));
```
antoniolinhart, line 87: Very clean update to the doPut!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -26,6 +27,7 @@
 import java.io.IOException;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
+import java.util.HashMap;
```
antoniolinhart, line 26: nit: unused imports

import com.google.cloud.firestore.QueryDocumentSnapshot;
import com.google.gson.Gson;
import com.google.gson.JsonElement;

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -54,21 +56,27 @@ public static JsonObject createRequestBodyJson(HttpServletRequest request) throw
...
+    if (id.length() == 0) {
+      System.err.println("Error caused by either an empty or non-existent \"id\" field in the post body.");
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+      return null;
+    }
     if (request.getParameterMap().size() > 1) {
       System.err.println("Error: No other parameter can be sent with an ID");
       response.sendError(HttpServletResponse.SC_BAD_REQUEST);
       return null;
     }
```
antoniolinhart: If you're now passing the id directly to the function, is it necessary to even pass the request as a parameter and to get its parameterMap? I'm guessing that you probably wouldn't be sending the id as a parameter and also as a field in the request.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -236,4 +244,76 @@ public static JsonObject createRequestBodyJson(HttpServletRequest request) throw
...
+          containsParameter = true;
+          if (type.equals("int") || type.equals("java.lang.Integer")) {
+            update.put(key, jsonObject.get(key).getAsInt());
+          }
+          if (type.equals("java.lang.String") && !field.getName().equals("id")) {
+            update.put(key, jsonObject.get(key).getAsString());
+          }
+          if (type.equals("boolean") || type.equals("java.lang.Boolean")) {
+            update.put(key, jsonObject.get(key).getAsBoolean());
+          }
```
antoniolinhart: nit: it appears these could be elseifs since type can't be an int and a boolean for instance

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -236,4 +244,76 @@ public static JsonObject createRequestBodyJson(HttpServletRequest request) throw
...
+
+    ApiFuture<WriteResult> writeResult = collectionReference.document(id).set(update, SetOptions.merge());
+    try {
+      System.out.println("Update time : " + writeResult.get().getUpdateTime());
+    } catch (Exception e) {
+      System.err.println("Error: " + e);
+      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+      return null;
+    }
+    return getById(id, collectionReference, request, response, genericClass).get(0);
```
antoniolinhart: Cool! I like how you're updating the data, then getting the updated object from the database again to return to the user

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
antoniolinhart: +1 to implementing this and Geoff's comment above which should make this a bit cleaner and remove this flag

antoniolinhart (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: Thanks!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Thanks for the catch! Once these last PRs are merged, I'm going to move over to gitpod, and it'll warn me about these sorts of things so you don't have to anymore :)

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Good call, I realized this in writing my post method. I used the else-ifs there, and also added a final else to catch any types that we do not support. I'll add that same functionality here.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: I could move the parameterMap check up into the get function, rather than leaving it here.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Thanks!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Good idea, I'll change this.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Will fix, definitely cleans it up.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Actually on second thought, I think it makes sense to loop through the JSON object. I want to grab everything in the request body, and use that to update the object. The only way to do this is by looping through all of its keys. Otherwise, I won't notice if they send keys that are not object fields. And, they do not necessarily need to send every field to be updated, which causes this to inappropriately throw errors. I think I'm going to revert back to looping through the key set.

MauroGuerrero20 (approved):
LGTM

Issue #9, opened 2020-06-18
BookList Search Integration


MauroGuerrero20:
### Features: 

- Allow users to add a book to their BookLists  using Search.

PR #46, opened 2020-07-07
[UI] Book Search Tile

antoniolinhart:
# Description
This PR focuses on the UI update of the BookSearchTile object, the view rendered on the Browse page after the User enters a search term and presses enter. It also sets the framework for the user to see lists of booklists they can add a book to.

# Breakdown
- Format the book information horizontally instead of being stacked as before
- Add a button for adding the given book to their booklist
  - Generates dynamically based on how many booklists the user has
- Add tests to incorporate the new `Add to List`  button
- Pass a "Book" object to the SearchBookTile instead of each primitive variable (authors, title, thumbnail..)

# Main Difficulties

Deciding which dropdown library to use - there are a lot of possibilities but I decided to use the ReactBootstrap components.

Setting up the test data to 'mock' the lists of BookLists I would get - I ended up just creating arrays with names and ids to show.

Ensuring that the dropdown was not covered by the buttons on the next lowest book element - with lots of help from Geoff, it now ensures that when the element is in focus, it will be the top layer seen.

# Screenshot of Feature

## Before this PR

Before Book Tile

## Created in this PR
This screenshot shows the tile view after hovering over one of the tiles with a lightblue background.
After Book Tile Hover

This screenshot shows the dropdown depicting the list of the user's clubs and communities that they can add the book to (currently dummy data).
After Book Dropdown

New tile is optimized for mobile view too!
After Book Mobile View

# Linked Issue
#9 

Creates Issue #48 for handling empty BookLists or Club/Communities


capstone-frontend/src/BookSearchTile.test.js
```diff
@@ -6,28 +6,55 @@ import { act } from "react-dom/test-utils";
...
+
+const getUserCommunities = () => {
+  let userCommunities = [{ id: "1", name: "Sci-fi Community" },
+  { id: "2", name: "Read-a-lot" }];
+  return userCommunities;
+}
 
 test('renders book search tile test', () => {
-  render(<BookSearchTile title={book.title} author={book.author} thumbnailLink={book.thumbnailLink} />, container);
+  render(<BookSearchTile book={book}
```
steven-solar, line 28: I like the way you turned the book into a json object instead of having several different props to pass!

capstone-frontend/src/BookSearchTile.test.js
```diff
@@ -6,28 +6,55 @@ import { act } from "react-dom/test-utils";
...
 
   const bookThumbnail = document.getElementsByClassName("book-img-med")[0];
   expect(bookThumbnail).toBeInTheDocument();
 
-  const bookTitle = document.getElementsByClassName("book-title")[0];
+  const bookTitle = document.getElementsByTagName("h2")[0];
   expect(bookTitle).toBeInTheDocument();
 
-  const bookAuthor = document.getElementsByClassName("book-author")[0];
+  const bookAuthor = document.getElementsByTagName("p")[0];
```
steven-solar: What was the reasoning for changing this to use tag names? Just worried in the case that we add other tags and this breaks. What would you think of giving them ids and selecting with those?

capstone-frontend/src/components/BookSearchList.jsx
```diff
@@ -7,26 +7,51 @@ class BookSearchList extends Component {
...
+    return userClubs;
+  }
+
+  getUserCommunities = () => {
+    let userCommunities = [{ id: "1", name: "Sci-fi Community" },
+    { id: "2", name: "Read-a-lot" }];
+    return userCommunities;
+  }
+
   componentDidMount() {
```
steven-solar, line 28: This structure is great! Will be super easy to swap out your fake data for fetch requests in the future. This is all organized really well.

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -1,14 +1,57 @@
...
 
 const BookSearchTile = (props) => {
   return (
     <div className="book-search-tile">
-      <img className="book-img-med" src={props.thumbnailLink} alt={props.title}/>
-      <div>
-        <div className="book-title"> {props.title} </div>
-        <div className="book-author"> {props.author.join()} </div>
-      </div>
+      <Container>
```
steven-solar, line 14: These tiles look great!

steven-solar (approved):

capstone-frontend/src/BookSearchTile.test.js
antoniolinhart: I removed the class names/ids for the title and author element because they currently aren't referenced in the CSS or elsewhere in the code - if it's valid to add an id to the element just so that I can test easier I'll definitely implement that. Thoughts @geoffbass ?

antoniolinhart: Discussed offline and I have now added class names to the elements and referred to them by class name in the test!

capstone-frontend/src/components/BookSearchList.jsx
antoniolinhart: Thank you!

capstone-frontend/src/components/BookSearchTile.jsx
antoniolinhart: Thanks! Hopefully we can use a similar format for the booklist search tile/club search tile when we get to implementing that in search!

capstone-frontend/src/BookSearchTile.test.js
```diff
@@ -6,28 +6,55 @@ import { act } from "react-dom/test-utils";
...
 const container = document.createElement("div");
 document.body.appendChild(container);
 const book = {
   title: "Title",
-  author: ["Author"],
+  authors: ["Author"],
   thumbnailLink: "https://http.cat/100.jpg"
 };
+const getUserBookLists = () => {
+  let userBookLists = [{ id: "1", name: "Best Books" },
```
MauroGuerrero20: Is this test separate from the main Component? Will the test break when I merge my PR?

MauroGuerrero20 (approved):
LGTM

capstone-frontend/src/BookSearchTile.test.js
antoniolinhart: Yeah, these tests are completely separate from the BookSearchList component. In this test, I just have these functions to make it easier to generate consistent test data.

VivekShankar (approved):
LGTM - really awesome work here Antonio! Love how the new tiles look, both in web and mobile views.

capstone-frontend/src/BookSearchTile.test.js
```diff
@@ -6,28 +6,55 @@ import { act } from "react-dom/test-utils";
...
+  { id: "2", name: "Fantasy" },
+  { id: "3", name: "Sci-fi" }];
+  return userBookLists;
```
geoffbass: nit: This would probably be a bit cleaner if it just directly returned the array literal instead of first storing it in a variable. If you decide to keep the variable, use `const` since it isn't reassigned.

Same applies to the helper functions below.

capstone-frontend/src/components/BookSearchList.jsx
```diff
@@ -7,26 +7,51 @@ class BookSearchList extends Component {
   constructor(props) {
     super(props);
     this.state = {
-      bookList: []
+      bookSearchList: []
```
geoffbass: nit: This will store a list of books that come back from the search right? In that case I think `bookList` or just `books` actually makes more sense here, particularly since `BookSearchList` is already the name of the component.

capstone-frontend/src/components/BookSearchList.jsx
```diff
@@ -7,26 +7,51 @@ class BookSearchList extends Component {
...
     }
-    this.getData = this.getData.bind(this);
   }
 
-  getData() {
+  getData = () => {
     fetch(`/api/search?searchTerm=${this.props.searchQuery}`)
       .then(response => response.json())
-      .then(res => this.setState({ bookList: res }))
+      .then(res => this.setState({ bookSearchList: res }))
```
geoffbass: nit: `res` here is actually the body/payload of the response I think, so calling it something like `books` would probably be more accurate.

capstone-frontend/src/components/BookSearchList.jsx
```diff
@@ -7,26 +7,51 @@ class BookSearchList extends Component {
...
       .catch(err => console.log(err));
   }
 
+  getUserBookLists = () => {
```
geoffbass, line 21: nit: Same as the helper functions I commented on in BookSearchTileTest - return the object literal directly for these.

I would also add a TODO with a linked issue number to replace these with actual API calls once the API is implemented.

capstone-frontend/src/components/BookSearchList.jsx
```diff
@@ -7,26 +7,51 @@ class BookSearchList extends Component {
...
+    let userCommunities = [{ id: "1", name: "Sci-fi Community" },
+    { id: "2", name: "Read-a-lot" }];
+    return userCommunities;
+  }
+
   componentDidMount() {
     this.getData();
+    this.setState({ userBookLists: this.getUserBookLists() });
+    this.setState({ userClubs: this.getUserClubs() });
+    this.setState({ userCommunities: this.getUserCommunities() });
```
geoffbass: nit: This can be done in a single call to `setState`.

capstone-frontend/src/components/BookSearchList.jsx
```diff
@@ -7,26 +7,51 @@ class BookSearchList extends Component {
...
   }
 
   render() {
     var books = [];
-    for (const book of this.state.bookList) {
-      books.push(<BookSearchTile title={book.title} author={book.authors} thumbnailLink={book.thumbnailLink} key={book.id} />);
+    for (const book of this.state.bookSearchList) {
```
geoffbass: nit: Rather than building the output list in a loop and then returning it afterward, you could return the result of `map`ping over `bookSearchList` to the `BookSearchTile`s directly here.

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -1,14 +1,57 @@
...
+              <h2 className="book-title"> {props.book.title} </h2>
+              <p className="book-authors"> {props.book.authors.join(', ')} </p>
+            </div>
+          </Col>
+          <Col md="auto">
+            <Container className="center-vertical">
+            <DropdownButton id="dropdown-list-add" className="dropdown-add" title="Add to List">
+              {
+                props.userBookLists.map(bookList => 
+                  React.createElement(Dropdown.Item, { key:bookList.id,
```
geoffbass: You should be able to use JSX here (e.g. `<Dropdown.Item key=... />` rather than `React.createElement`.

capstone-frontend/src/styles/BookSearchTile.css
```diff
@@ -0,0 +1,30 @@
...
+  z-index: 100;
+}
+
+.dropdown-add {
+  margin-bottom: 5px;
+}
+
+#dropdown-group-add, #dropdown-list-add {
+  width: 100%;
+  position: relative;
```
geoffbass: nit: Alphabetize declarations

geoffbass (approved):
LGTM, just some small changes to make before you merge. Let me know if you have questions.

capstone-frontend/src/BookSearchTile.test.js
antoniolinhart: That's a good idea! I've fixed that now

capstone-frontend/src/components/BookSearchList.jsx
antoniolinhart: I think I'll use `books`. Mauro and I thought that `bookList` was ambiguous since he will be implementing a list of `bookLists` in this state as well so deciphering between bookList and bookLists (or similar) would get confusing.

capstone-frontend/src/components/BookSearchList.jsx
antoniolinhart: Fixed!

capstone-frontend/src/components/BookSearchList.jsx
antoniolinhart: Removed userClubs and userCommunities from this for now, but will keep this in mind.

capstone-frontend/src/components/BookSearchList.jsx
antoniolinhart: Added!

capstone-frontend/src/styles/BookSearchTile.css
antoniolinhart: I alphabetized the classes, just not those declarations!

capstone-frontend/src/components/BookSearchList.jsx
antoniolinhart: Good idea! I didn't know it was possible to do it this way until I worked on it in the BookSearchTile

capstone-frontend/src/components/BookSearchTile.jsx
antoniolinhart: Changed! I thought there must be a cleaner way to do it than using createElement

capstone-frontend/src/styles/BookSearchTile.css
antoniolinhart: Fixed!

PR #51, opened 2020-07-08
Abstract post

### LINKED ISSUE #35: Generic GET/POST/PUT Database Access

steven-solar:
This PR creates a generalized put method in the Utility class for all servlets to use in the future. It assures that all provided fields in the request body are fields of the object to be created (names must match exactly). Fields can be of type int/Integer, boolean/Boolean, String or List<String>. These supported types can be expanded at a later date if necessary. It also allows the servlet to supply certain "required fields" that must be present, and non-empty, in order for the object to be "valid." If any of these fields are missing, errors will be thrown. IDs are generated automatically by firebase. The actual object in the firebase does not have an id field, but the returned object contains the id in its `id` field. The function will return the created object, which will then be put into the response body. In order to do this, the getById helper function is used.

All servlets can call the generalized post method to create any field of any object class. The method performs all necessary error handling (on errors, bad requests, or when an id is not found), the method will attach response headers of 400, 404 and 500 appropriately, and return a null object). Servlets should call this method, and print its results to the HttpServletResponse response.

This PR also makes slight updates to the CreateClub component, allowing it to post memberIDs, and removing the user-inputted id (as it is now generated by firebase).


capstone-frontend/src/components/CreateClub.jsx
```diff
@@ -13,15 +13,13 @@ class CreateClub extends Component {
...
-          console.log(e);
-          alert("Looks like we're having trouble connecting to our database, hang tight!");
-        });  
+    let response = fetch("/api/clubs", {method: "post", body: JSON.stringify(data)});
+    response.catch(function(err) {
+      alert(err);
+    });
+    response.then(response => response.json()).then(response => {
+      history.push(`/clubpage/${response.id}`);
+    });
```
antoniolinhart: Is there a way to surface this alert in the case of error without splitting it into a response variable then catching the error before you convert the json and push the new link?

If the code is invalid JSON then won't it throw an error that you could catch as in your original code and not continue to the `history.push`?

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -316,4 +319,114 @@ else if (("remove_"+field.getName()).equals(key)) {
...
+                    .filter(key -> jsonObject.get(key).getAsString().length() > 0)
+                    .collect(Collectors.toList()));
+    list.retainAll(requiredFields);
+    if (list.size() != requiredFields.size()) {
+      System.err.println("Error: Not all required fields are present in the request body, or some required fields are empty.");
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+      return false;
+    } 
+    return true;
+  }
```
antoniolinhart: I think this could use a couple comments around each section of code in this. I know your System.err.println messages roughly indicate what it's supposed to do but it could be written out a bit more

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -316,4 +319,114 @@ else if (("remove_"+field.getName()).equals(key)) {
...
+        else {
+          constructorFields.put(name, -1);
+        }
+      }
+      else if (type.equals("java.lang.String")) {
+        if (jsonObject.has(name) && jsonObject.get(name).getAsString().length() != 0) {
+          constructorFields.put(name, jsonObject.get(name).getAsString());
+        }
+        else {
+          constructorFields.put(name, "");
```
antoniolinhart: I like your use of these "default" values for each datatype.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -316,4 +319,114 @@ else if (("remove_"+field.getName()).equals(key)) {
...
+      HttpServletResponse response, GenericClass<T> genericClass, List<String> requiredFields) throws IOException {
+    Map<String, Object> constructorFields = new HashMap<>();
+    JsonObject jsonObject = Utility.createRequestBodyJson(request);
+    Field[] fields = genericClass.getMyType().getDeclaredFields();
+    if (!postErrorHandler(jsonObject, response, fields, requiredFields)) {
+      return null;
+    }
+    for (Field f : fields) {
+      String name = f.getName();
+      String type = f.getType().getName();
```
antoniolinhart: nit: It might be more clear if you change this to fieldName or fName? "name" makes me think it's a field of an object, like the name of a BookList or the name of a User

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -316,4 +319,114 @@ else if (("remove_"+field.getName()).equals(key)) {
...
+      }
+      else if (type.equals("java.lang.String")) {
+        if (jsonObject.has(name) && jsonObject.get(name).getAsString().length() != 0) {
+          constructorFields.put(name, jsonObject.get(name).getAsString());
+        }
+        else {
+          constructorFields.put(name, "");
+        }
+      }
+      else if (type.equals("boolean") || type.equals("java.lang.boolean")) {
```
antoniolinhart, line 425: This is just an idea, but maybe this could be replaced with type.contains("boolean") which will match either one of "boolean" or "java.lang.boolean". If you're worried about casing, you could do type.toLowerCase().contains("boolean")

Applies to your other type checks as well.

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -52,31 +53,12 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-
-    Club club = new Club(name, description, ownerID, gbookID);
-    ApiFuture<WriteResult> asyncDocument = clubs.document(id).set(club);
-    try {
-      System.out.println("Update time : " + asyncDocument.get().getUpdateTime());
-    } catch (Exception e) {
-      System.err.println("Error: " + e);
-      response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
-      return;
+    List<String> requiredFields = new ArrayList<String>( Arrays.asList("name", "ownerID") );
```
MauroGuerrero20, line 56: Is there a way to maybe not have to require the fields in a ```List```

antoniolinhart (approved):
LGTM, just a few comments

MauroGuerrero20 (commented):
Hi Steven, I really don't understand you code. Is it possible to maybe setup a GVC with Antonio and I.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Thanks!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Makes sense, I'll change it for readability.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Sure I'll add.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: I could do that, but I think I want to leave it as is to make it clear how I'm checking. I think using the contains() could be more confusing than explicitly checking for both the primitive and Object type.

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: These fields are only the ones deemed "required" for object creation. For example, if you were creating a Club you could do without passing it a gbookID or memberIDs when you first create it. But maybe you want to force inclusion of an ownerID and a name on creation. That's what the requiredFields lets you accomplish. The method will still check all the fields you pass and use them in object creation, but it will throw errors if the requiredFields are not all present in the request body.

capstone-frontend/src/components/CreateClub.jsx
steven-solar: I was a little confused as to how to set this up, as I wanted to catch errors on the fetch as well as the later conversion to json. Maybe adding another .catch() to the promise chain where I convert to json and push to history would cover the case you mentioned.

MauroGuerrero20 (approved):
LGTM

capstone-frontend/src/components/CreateClub.jsx
steven-solar: After re-visiting the old code that I replaced, I think that did the trick better than this. I'll revert these changes.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -349,4 +352,116 @@ else if (key.equals("remove_"+name)) {
...
+    list.addAll(jsonObject.keySet());
+    list.retainAll(fieldNames);
+    // Every jsonObject key must match a field name. If not, this check will fail. 
+    if (list.size() != jsonObject.keySet().size()) {
+      System.err.println("Error: Not all parameters in the request body are fields of the given class.");
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+      return false;
+    } 
+    list.clear();
+    list.addAll(jsonObject.keySet().stream()
```
VivekShankar, line 377: Is there a reason you're filtering out all "empty string" keys from `jsonObject`? Wouldn't `list.retainAll(requiredFields)` work even if `list` contained empty strings?

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -349,4 +352,116 @@ else if (key.equals("remove_"+name)) {
...
+   * @return boolean true if it is a valid post request, false if not
+   */
+  public static <T extends BaseEntity> boolean postErrorHandler(JsonObject jsonObject, HttpServletResponse response, 
+      Field[] fields, List<String> requiredFields) throws IOException {
+    List<String> list = new ArrayList<>();
+    List<String> fieldNames = Arrays.asList(fields).stream().map(f -> f.getName()).collect(Collectors.toList());
+    list.addAll(jsonObject.keySet());
+    list.retainAll(fieldNames);
+    // Every jsonObject key must match a field name. If not, this check will fail. 
+    if (list.size() != jsonObject.keySet().size()) {
```
VivekShankar: I'm wondering if you can simply use the `containsAll` method to simplify the `addAll`, `retainAll`, size-check combination. For example, would `fieldNames.containsAll(jsonObject.keySet())` work in the first check?

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: I filter out empty keys to make sure that the required fields are "actually there" and not just value-less keys in the json. For example, if an `ownerID` is deemed necessary for Club creation, then an empty String ownerID would not suffice and the post should fail in this case.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -349,4 +352,116 @@ else if (key.equals("remove_"+name)) {
...
+        else {
+          constructorFields.put(fName, false);
+        }
+      }
+      else if (type.equals("java.util.List") || type.equals("java.util.ArrayList")) {
+        if (jsonObject.has(fName) && jsonObject.get(fName).getAsString().length() != 0) {
+          // First check to see if it's an "actual array"
+          try {
+            JsonElement value = jsonObject.get(fName);
+            Type listType = new TypeToken<List<String>>() {}.getType();
```
VivekShankar, line 438: Nice job figuring out how to parse a JsonElement into a list!

VivekShankar (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
VivekShankar: Ah, thanks! I missed that you were filtering out keys with **values** that are empty. Can you add a quick update to the comment: "Every requiredField must be present as a key with a **non-empty** value..."?

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: I think that would work in this case. Here I'm trying to check that every `jsonObject` key matches a `fieldName`, and I think `containsAll` would do that in a cleaner and more obvious way.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Thanks!

PR #55, opened 2020-07-09
Generalized delete helper function

### LINKED ISSUE #35: Generic GET/POST/PUT Database Access

steven-solar:
This PR created a `delete` function servlets can call to delete objects by id. This will work for all servlets, and return true if the object is deleted successfully and false if not. The PR also creates a delete button on the clubpage to use the new functionality. The completion of this function means that all basic CRUD operations now have been added to the Utility class for servlets to use.


MauroGuerrero20 (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
```diff
@@ -69,4 +69,11 @@ public void doPut(HttpServletRequest request, HttpServletResponse response) thro
       response.getWriter().println(gson.toJson(updatedClub));
     }
   }
+
+  @Override
+  public void doDelete(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    boolean successfulDelete = Utility.delete(clubs, request, response);
+    response.setContentType("application/json;");
+    response.getWriter().println(gson.toJson(successfulDelete));
```
geoffbass: Will this send a 200 response with a payload `true` when the delete is successful? I'd expect to see a 204 (No Content) without a payload on success.

I see you're handling errors and sending the appropriate responses in the utility function, nice work! I think it would also be helpful to wrap lines 76 and 77 here in a conditional on `successfulDelete` to make it clear this is only a success path and to avoid continuing to modify the response object after it's sent (though the servlet framework might handle that, I'm not sure).

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -14,6 +14,13 @@ class ClubPage extends Component {
         .then(response => response.json()).then(res => this.setState({club: res[0]}));
   }
 
+  handleClick = (e) => {
+    e.preventDefault();
+    fetch(`/api/clubs?id=${this.props.match.params.id}`, {method: "delete"})
+    .then(response => {console.log(response)});
```
geoffbass: This should also handle errors (`.catch`). Consider what you want the front-end to do when the delete fails on the back-end: surface a failure message to the user in the UI? automatically retry the request a certain number of times (bonus points for exponential backoff but not necessary at this stage)? Regardless, you'll want to set up some sort of client-side error logging at some point (where the errors are sent to a centralized place you have access to, not just logged in the client console).

Logging the error in the console and putting a TODO (with an associated issue) here to handle errors properly is fine at this point, just make sure you're thinking about this and document the issue.

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -14,6 +14,13 @@ class ClubPage extends Component {
...
+    this.props.history.push("/myclubs");
```
geoffbass: This should probably happen only once you know the delete was successful, though that change should also be paired with some sort of user-facing error message when it was unsuccessful.

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -14,6 +14,13 @@ class ClubPage extends Component {
         .then(response => response.json()).then(res => this.setState({club: res[0]}));
   }
 
+  handleClick = (e) => {
+    e.preventDefault();
```
geoffbass: Not sure this is necessary here. `preventDefault` is usually needed in methods that handle form submissions, since the default submit behavior is to redirect the browser. I don't think there's any click-handling default behavior you need to suppress though.

geoffbass (approved):
LGTM with a few small edits. Reach out if you have follow-up questions!

capstone-frontend/src/components/ClubPage.jsx
antoniolinhart: I looked up frontend logging on GCP and it looks like Winston may be something we can use to send errors similar to the way we send the backend Java errors to the console?
https://cloud.google.com/logging/docs/setup/nodejs#using_winston

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -464,4 +464,42 @@ else if (type.equals("java.util.List") || type.equals("java.util.ArrayList")) {
...
+    if (request.getParameter("id") != null) {
+      String id = request.getParameter("id");
+      if (id.length() == 0) {
+        System.err.println("Error caused by an empty \"id\" field in the post body.");
+        response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+        return false;
+      }
+      ApiFuture<WriteResult> writeResult = collectionReference.document(id).delete();
+      try {
+        System.out.println("Update time : " + writeResult.get().getUpdateTime());
```
antoniolinhart: I looked at this similar update code in Mauro's PR which he changed to have the name of the method, would it be good to have `"Delete Update Time :"` to be a little more specific? Not sure whether the backend logs provide the method signature

antoniolinhart (approved):
LGTM! Really cool to have all of these generic methods we can use for database access now

capstone-backend/src/main/java/com/google/sps/servlets/ClubServlet.java
steven-solar: As written it will, but I'll change it to send a `204` instead. In the same vein, I'll remove lines 76 and 77 as now all handling is done in the Utility class. If the object is successfully deleted, Utility will send a 204. If it is not, some error code will be sent.

capstone-frontend/src/components/ClubPage.jsx
steven-solar: This is a good thing to think about, and we'll discuss as a team how we want to handle this. In the meantime, I created an issue and marked the TODO here (and below). I also now only push the user to the homepage in the case of a successful delete.

capstone-frontend/src/components/ClubPage.jsx
steven-solar: I became so used to adding it to all my handlers, I didn't realize it was unneeded here. Will remove.

capstone-frontend/src/components/ClubPage.jsx
steven-solar: Will move the history push into the `then` clause and add a `catch` as discussed above. For now we'll simply use a javascript alert to signify a failed delete. In the future the exponential backoff retry would be a great solution to the problem.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: That's a good idea, and could be helpful in debugging later on. I'll change these print statements in the other functions as well.

PR #56, opened 2020-07-09
Add conditional button rendering to BookSearchTile Add

### LINKED ISSUE #48: BookSearchTile button update
When there are no BookLists to show on the BookSearchTile, display a button with a different color that directs the user to **create** a new BookList

Update: removed references to Club lists and Communities lists as these will be added in other components

antoniolinhart:
# Description

This PR focuses on adding a condition to which button is rendered by the BookListAdd button on the BookSearchTile. If there are no BookLists from the user, it will give the user the option to go to a page to create a new BookList. The button shown in PR #46 will be shown if there are 1 or more BookLists to add books to.

# Breakdown

- Add new sub components for both cases to render different dropdown menus
  - BookListDropdownActive (1 or more booklists)  
  - BookListDropdownEmpty (no booklists)
    - Clicking on this button directs user to the CreateList page

# Main Difficulties

Deciding how to abstract out the two different buttons - whether to keep them in the ternary, or move into separate functional components.
Then, deciding what to call these buttons. Open to feedback if they are unclear or share too much code.

# Screenshot of Feature

Display when there are no BookLists with the current user's UserId.

Add to List Button No BookLists

# Linked Issue

#48


capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -22,16 +21,11 @@ const BookSearchTile = (props) => {
...
-                      onSelect={() => addBookToBookList(props.book.id, bookList.id)}>
-                      {bookList.name}
-                    </Dropdown.Item>
-                  )
-                }
-              </DropdownButton>
+              {props.userBookLists.length > 0 ?
+                <BookListDropdownActive userBookLists={props.userBookLists} book={props.book}></BookListDropdownActive>
+                :
+                <BookListDropdownEmpty history={props.history}></BookListDropdownEmpty>
```
antoniolinhart: History isn't needed as a prop here

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -40,4 +34,31 @@ const BookSearchTile = (props) => {
...
+      }
+    </DropdownButton>
+  );
+}
+
+const BookListDropdownEmpty = (props) => {
+  return (
+    <DropdownButton id="dropdown-list-add" className="dropdown-add" title="No Lists Found" variant="warning">
+      {
+        <Dropdown.Item href="/createlist" key={"emptyBookList"}>
```
antoniolinhart: These `{ }` aren't needed around emptyBookList

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -40,4 +34,31 @@ const BookSearchTile = (props) => {
   );
 }
 
+const BookListDropdownActive = (props) => {
```
steven-solar: Would it be possible to make one BookListDropdown component that renders conditionally based on the length of userBookLists? We could do the conditional rendering inside the BookListDropdown instead of outside the way you do it here. Not necessary but I think it would be cleaner.

steven-solar (approved):
Looks great!

capstone-frontend/src/components/BookSearchTile.jsx
antoniolinhart: Fixed

capstone-frontend/src/components/BookSearchTile.jsx
antoniolinhart: Fixed!

capstone-frontend/src/components/BookSearchTile.jsx
antoniolinhart: Yup! This is a good idea. It also makes it less confusing to define two different functions with different names. Changed and confirmed it still works as expected!

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -8,12 +8,11 @@ const addBookToBookList = (bookId, bookListId) => {
 }
 
 const BookSearchTile = (props) => {
-  // TODO: #48 Prompt user to create a BookList if they have no BookLists
   return (
     <div className="book-search-tile">
       <Container>
         <Row className="justify-content-md-center">
-          <Col md="auto"><img className="book-img-med" src={props.book.thumbnailLink} alt={props.book.title}/></Col>
+          <Col md="auto"><img className="book-img-med" src={props.book.thumbnailLink} alt={props.book.title} /></Col>
```
geoffbass: nit: You should be able to omit the closing `</Col>` tag since the tag is self-closing with the `/` at the end. Agree with adding the extra space here though.

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -22,16 +21,7 @@ const BookSearchTile = (props) => {
...
-                {
-                  props.userBookLists.map(bookList =>
-                    <Dropdown.Item key={bookList.id}
-                      onSelect={() => addBookToBookList(props.book.id, bookList.id)}>
-                      {bookList.name}
-                    </Dropdown.Item>
-                  )
-                }
-              </DropdownButton>
+              <BookListAddDropdown userBookLists={props.userBookLists} book={props.book}></BookListAddDropdown>
```
geoffbass: nit: Same here - make this a self-closing tag if there are no children.

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -40,4 +30,31 @@ const BookSearchTile = (props) => {
   );
 }
 
+const BookListAddDropdown = (props) => {
```
geoffbass: (No action required) nit: Consider destructuring `props` here, i.e. `({ book, userBookLists })`. This makes the function signature more informative and allows you to omit `props.` in the JSX below.

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -40,4 +30,31 @@ const BookSearchTile = (props) => {
...
+            </Dropdown.Item>
+          )
+        }
+      </DropdownButton>
+    );
+  } else {
+    return (
+      <DropdownButton id="dropdown-list-add" className="dropdown-add" title="No Lists Found" variant="warning">
+        {
+          <Dropdown.Item href="/createlist" key="emptyBookList">
```
geoffbass: I don't think you need `key` here since this isn't a list item.

geoffbass (approved):
LGTM

MauroGuerrero20 (approved):
LGTM

capstone-frontend/src/components/BookSearchTile.jsx
antoniolinhart: When I remove the closing Col tag, it tells me there is a JSX syntax error. Could this be because the img tag is a child of this object so it's necessary?

capstone-frontend/src/components/BookSearchTile.jsx
antoniolinhart: Fixed!

capstone-frontend/src/components/BookSearchTile.jsx
antoniolinhart: Fixed!

capstone-frontend/src/components/BookSearchTile.jsx
antoniolinhart: This is a good idea! I didn't know about it before

capstone-frontend/src/components/BookSearchTile.jsx
geoffbass: My mistake! Totally missed the opening `<img` tag so this makes much more sense now. I'd move the `img` onto its own indented line so it's harder to miss but otherwise this looks good.

PR #58, opened 2020-07-10
Assignments and Comments: objects and servlets

steven-solar:
This PR creates the Assignment and Comment objects, and their corresponding servlets. It also modifies the Club object to contain a list of Assignments rather than posts. The Group object no longer contains a list of posts, as this has been moved to be a Community-only feature. Additionally, the announcement field was removed from these objects. For clubs, all discussion will take place in the form of Comments underneath Assignments. The appropriate testing is updated.

Additionally, now that Club and Communities objects extend the Group class, the use of getDeclaredFields on the given class was insufficient as it does not get fields of super classes. This PR also updates the Utility class to get the fields of the superclass as well, solving this problem.


capstone-backend/src/main/java/com/google/sps/data/Assignment.java
```diff
@@ -0,0 +1,69 @@
...
+
+  public String getClubID() {
+    return clubID;
+  }
+
+  public List<Comment> getComments() {
+    return comments;
+  }
+
+  public void setComments(ArrayList<Comment> comments) {
```
MauroGuerrero20, line 46: I still don't know why we need setters for all objects? Do you know where in the Firestore documentation its say they are required?

capstone-backend/src/main/java/com/google/sps/data/Comment.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class Comment extends BaseEntity {
+  private final String text;
+  private final String userID;
+  private final String whenCreated;
+
+  public Comment(String text, String userID, String whenCreated) {
+    this.text = text;
+    this.userID = userID;
+    this.whenCreated = whenCreated;
```
MauroGuerrero20, line 12: Will this Comment object be use to store comments on Firestore? Don't forget to take a look into Firebase Messaging API.

- https://firebase.google.com/docs/cloud-messaging/js/client
- https://firebase.google.com/docs/cloud-messaging/js/receive
- https://www.youtube.com/watch?v=BsCBCudx58g

capstone-backend/src/test/java/com/google/sps/data/ClubTest.java
```diff
@@ -10,7 +10,19 @@
...
+  private static final String ORIGINAL_DESCRIPTION = "originalDescription";
+  private static final String NEW_DESCRIPTION = "newDescription";
+
+  private static final String ANNOUNCEMENT = "announcement";
+
+  private static final Assignment ASSIGNMENT_ONE = new Assignment("one", "id", "created", "due");
+  private static final Assignment ASSIGNMENT_TWO = new Assignment("two", "id", "created", "due");
+
+  private static final String ORIGINAL_OWNER = "originalOwner";
+  private static final String NEW_OWNER = "newOwner";
```
MauroGuerrero20:
```
ORIGINAL_NAME
ORIGINAL_DESCRIPTION
ORIGINAL_OWNER
```
Are never used. Consider removing them.

capstone-backend/src/test/java/com/google/sps/data/ClubTest.java
```diff
@@ -28,18 +40,178 @@ public void setUp() {
...
+  public void testInvite() {
+    club.invite(MEMBER_ONE);
+
+    Assert.assertEquals(club.getInviteIDs().size(), 1);
+    List<String> list = club.getInviteIDs();
+    Assert.assertTrue(list.contains(MEMBER_ONE));
+
+    club.invite(MEMBER_TWO);
+
+    Assert.assertEquals(club.getInviteIDs().size(), 2);
```
MauroGuerrero20: Gitpod is showing a lot of errors. I am guessing this tests aren't done.

MauroGuerrero20 (commented):
I am guessing the PR still needs work. Some of the tests failed on Gitpod.

capstone-backend/src/main/java/com/google/sps/data/Assignment.java
steven-solar: If you don't include getters and setters for objects, Firestore will throw errors that object fields "cannot be serialized." While they are not explicitly called in our code, Firestore uses them to create and update object fields and we need to include them.

capstone-backend/src/test/java/com/google/sps/data/ClubTest.java
steven-solar: Looks like last commit didn't push, will push and resolve this now.

capstone-backend/src/test/java/com/google/sps/data/ClubTest.java
steven-solar: Similar to below comment, pulled this out in the last commit which didn't push. Thanks for catching this.

capstone-backend/src/main/java/com/google/sps/data/Comment.java
steven-solar: I wanted to discuss this with the team, as we try to figure out what we should fit into our MVP and what is not core to our application. I looked into the API, and we can discuss as a team on Monday.

capstone-backend/src/main/java/com/google/sps/servlets/AssignmentServlet.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }
+
+  @Override
+  public void doPut(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Assignment updatedAssignment = (Assignment) Utility.put(assignments, request, response, new GenericClass(Assignment.class));
+    if (updatedAssignment != null) {
+      response.setContentType("application/json;");
+      response.getWriter().println(gson.toJson(updatedAssignment));
+    }
```
antoniolinhart, line 57: Very clean get/post/put!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -132,7 +132,9 @@ public static JsonObject createRequestBodyJson(HttpServletRequest request) throw
       return null;
     }
     String parameterValue = entry.getValue()[0];
-    Field[] fields = genericClass.getMyType().getDeclaredFields();
+    Class superClass = genericClass.getMyType().getSuperclass();
+    List<Field> fields = new ArrayList<>(Arrays.asList(genericClass.getMyType().getDeclaredFields()));
+    fields.addAll(Arrays.asList(superClass.getDeclaredFields()));
```
antoniolinhart, line 137: I'm glad this get superclass declared fields works!

capstone-backend/src/test/java/com/google/sps/CommunityTest.java
```diff
@@ -33,7 +36,6 @@ public void setUp() {
   public void testConstructor() {
     Assert.assertEquals(community.getName(), NAME);
     Assert.assertEquals(community.getDescription(), DESCRIPTION);
-    Assert.assertEquals(community.getAnnouncement(), "");
```
antoniolinhart: I think this was the line that was causing the test to fail that Mauro is talking about? This test is failing on my machine but the getAnnouncement() line it is talking about isn't there, so I think it's fine.

antoniolinhart (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -132,7 +132,9 @@ public static JsonObject createRequestBodyJson(HttpServletRequest request) throw
       return null;
     }
     String parameterValue = entry.getValue()[0];
-    Field[] fields = genericClass.getMyType().getDeclaredFields();
+    Class superClass = genericClass.getMyType().getSuperclass();
+    List<Field> fields = new ArrayList<>(Arrays.asList(genericClass.getMyType().getDeclaredFields()));
```
VivekShankar, line 136: Have you tried testing this with a class that doesn't have an explicit super-class?  Although, I think your implementation should work: I believe `superClass` will be `java.lang.Object` and `superClass.getDeclaredFields()` should return an empty list.

capstone-backend/src/main/java/com/google/sps/servlets/AssignmentServlet.java
```diff
@@ -0,0 +1,59 @@
...
+    List<Assignment> retrievedAssignments = Utility.get(assignments, request, response, new GenericClass(Assignment.class));
+    if (retrievedAssignments != null) {
+      response.setContentType("application/json;");
+      response.getWriter().println(gson.toJson(retrievedAssignments));
+    }
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    List<String> requiredFields = new ArrayList<String>( Arrays.asList("text", "clubID", "whenCreated") );
```
VivekShankar, line 43: Just curious... Is `whenDue` not a required field for an `Assignment` - i.e., can users create assignments without an explicit due date?

VivekShankar (approved):
LGTM - nice work!

capstone-backend/src/main/java/com/google/sps/servlets/AssignmentServlet.java
steven-solar: Thanks!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: Me too! Lets us deal with our Group hierarchy nicely.

steven-solar: If we have "deeper" hierarchies, this may fail. Doesn't seem to be an issue for now but worth looking into in the future.

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
steven-solar: I haven't tested it since in our case all classes at the very least extend BaseEntity (for ids). In general, I believe what you're saying is correct, since the `superClass` would be `java.lang.Object` which has no fields. The documentation says "If this Class object represents a class or interface with no declared fields, then this method returns an array of length 0." (https://docs.oracle.com/javase/10/docs/api/java/lang/Class.html#getDeclaredFields()) So in this case, appending the empty array wouldn't change anything, which is what we want. I could add an explicit check, but your reasoning was why I didn't.

capstone-backend/src/main/java/com/google/sps/servlets/AssignmentServlet.java
steven-solar: That was the idea. I thought it could be nice for open discussion that isn't necessarily "due."

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -365,9 +371,9 @@ else if (key.equals("remove_"+name)) {
    * @return boolean true if it is a valid post request, false if not
    */
   public static <T extends BaseEntity> boolean postErrorHandler(JsonObject jsonObject, HttpServletResponse response, 
-      Field[] fields, List<String> requiredFields) throws IOException {
+      List<Field> fields, List<String> requiredFields) throws IOException {
     List<String> list = new ArrayList<>();
-    List<String> fieldNames = Arrays.asList(fields).stream().map(f -> f.getName()).collect(Collectors.toList());
+    List<String> fieldNames = fields.stream().map(f -> f.getName()).collect(Collectors.toList());
     // Every jsonObject key must match a field name. If not, this check will fail. 
     if (fieldNames.containsAll(jsonObject.keySet())) {
```
MauroGuerrero20, line 378: Add '!'

MauroGuerrero20 (approved):
LGTM. Just add '!' on that one line we debugged

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
VivekShankar: Sounds great, thanks for exploring the docs and clarifying the expected behavior here!

PR #60, opened 2020-07-13
Update Mockito

antoniolinhart:
# Description
Update Mockito version to 2.22.0

Each time the tests were running it was reporting an illegal refraction, so I updated the mockito-core version recommended online and now the tests run as normal!

Factory path updated automatically when I rebuilt using maven.


steven-solar (approved):
Nice fix!

MauroGuerrero20 (approved):
New Tech Mockito

VivekShankar (approved):
LGTM

PR #62, opened 2020-07-13
Search Servlet Add Search by Id

### LINKED ISSUE #63: Search Servlet Search by Id
To be able to have an individual book page and load extra data not shown in the BookSearchTile, there needs to be another case for accessing the `doGet` of the `SearchServlet` by passing the `gbookId` instead of a `searchTerm`

antoniolinhart:
# Description

This PR focuses on adding a case to the SearchServlet when there is a `gbookId` parameter. When there is a `gbookId` parameter, it will retrieve a List of size 1 containing the volume with that id.

This means the two separate cases are
1. gbookId
2. searchTerm and maxResults (optional)

If neither of these are passed, it will return a SC_BAD_REQUEST.

# Breakdown

- Separate doGet to search for 2 cases mentioned above
- Abstract convertResponseToVolumeData into two methods so there is a way to parse the singular book retrieved from the Google Books individual volume call
- Tests
  - Add 3 new tests for the individual response and new error codes
  - Separate artificial JSON response into 3 separate strings to be able to used for the individual book test
  - Minor method to allow for a doGet with improper parameters

# Main Difficulties

Deciding how much of the code in the doGet to keep "global" and what to maintain only in certain if suites. For example, I decided to keep the formatted URL as a "shared" variable since its value was needed and would be modified depending on the value of the parameters.

Returning a list of the VolumeData object(s) no matter whether there is a single VolumeData object or many, I took inspiration from @steven-solar in his abstract get in the Utility class.

# Linked Issue

#63

### DIFF
--- a/capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
@@ -59,7 +59,7 @@ public class SearchServlet extends HttpServlet {
       formattedURL = String.format("https://www.googleapis.com/books/v1/volumes/%s", gbookId);
 
     } else if (request.getParameter("searchTerm") != null || !request.getParameter("searchTerm").isEmpty()) {
-      if (request.getParameterMap().size() >= 2 && request.getParameter("maxResults") != null) {
+      if (request.getParameterMap().size() >= 2 && request.getParameter("maxResults") == null) {
         System.err.println("Error: No other parameters must be sent with searchTerm other than maxResults");
         response.sendError(HttpServletResponse.SC_BAD_REQUEST);
         return;
--- a/capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
@@ -31,7 +31,6 @@ import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.stream.Collectors;
-
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
--- a/capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
@@ -58,7 +58,7 @@ public class SearchServlet extends HttpServlet {
       String gbookId = request.getParameter("gbookId");
       formattedURL = String.format("https://www.googleapis.com/books/v1/volumes/%s", gbookId);
 
-    } else if (request.getParameter("searchTerm") != null || request.getParameter("searchTerm").isEmpty()) {
+    } else if (request.getParameter("searchTerm") != null || !request.getParameter("searchTerm").isEmpty()) {
       String searchTerm = URLEncoder.encode(request.getParameter("searchTerm"), "UTF-8");
       // format url
       int maxResults = DEFAULT_MAX_RESULTS;
--- a/capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
@@ -59,6 +59,11 @@ public class SearchServlet extends HttpServlet {
       formattedURL = String.format("https://www.googleapis.com/books/v1/volumes/%s", gbookId);
 
     } else if (request.getParameter("searchTerm") != null || !request.getParameter("searchTerm").isEmpty()) {
+      if (request.getParameterMap().size() >= 2 && request.getParameter("maxResults") != null) {
+        System.err.println("Error: No other parameters must be sent with searchTerm other than maxResults");
+        response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+        return;
+      }
       String searchTerm = URLEncoder.encode(request.getParameter("searchTerm"), "UTF-8");
       // format url
       int maxResults = DEFAULT_MAX_RESULTS;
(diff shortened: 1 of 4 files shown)


capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -45,29 +45,42 @@
...
+      if (request.getParameterMap().size() > 1) {
+        System.err.println("Error: No other parameter can be sent with a gbookId");
+        response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+        return;
+      }
+      // make formatted url
+      String gbookId = request.getParameter("gbookId");
+      formattedURL = String.format("https://www.googleapis.com/books/v1/volumes/%s", gbookId);
+
+    } else if (request.getParameter("searchTerm") != null || request.getParameter("searchTerm").isEmpty()) {
```
VivekShankar: Just a quick clarification on this condition - I see that in your previous implementation, you're returning if either the `searchTerm` is `null` or `isEmpty`. Therefore, should we have this condition be the "negation" of that previous condition, i.e.,

`else if (request.getParameter("searchTerm") != null && !request.getParameter("searchTerm").isEmpty())`

capstone-backend/src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -100,23 +141,45 @@ public void ensureCustomNumberOfBooksInResponse() {
...
-    
-    boolean anyNotPopulated = books.stream()
-      .anyMatch( book -> book.getID() != null ||
-      book.getTitle() != null || 
-      book.getAuthors() != null ||
-      book.getDescription() != null ||
-      book.getThumbnailLink() != null);
+    Collection<VolumeData> books = SearchServlet.convertResponseToVolumeData(ARTIFICIAL_JSON_RESPONSE);
+
+    boolean anyNotPopulated = books.stream().anyMatch(book -> checkPopulatedVolumeDataObject(book));
```
VivekShankar: Should this be renamed to `anyPopulated`? Since you're using `anyMatch` it looks like this expression will be true as long as at least one book contains a non-null field. Alternatively, do you want this test to check that **at least one** book is populated, or that **all** books are populated?

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
antoniolinhart: Yes, this is correct! I just missed the ! in front of the second part of that statement

Thank you for catching this!

capstone-backend/src/test/java/com/google/sps/servlets/SearchServletTest.java
antoniolinhart: I changed this to `allPopulated` using `allMatch` instead of `anyMatch` I also refactored `checkPopulatedVolumeDataObject` to return false if there are any default values!

It was intended to check that **all** books are populated with nondefault values.

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -81,27 +94,42 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      }
+    } else if (request.getParameter("searchTerm") != null) {
+      volumes.addAll(convertResponseToVolumeData(fullOutput));
+      if (volumes.size() == 0) {
+        response.sendError(HttpServletResponse.SC_NOT_FOUND);
+        return;
+      }
+    }
+
+    Gson gson = new GsonBuilder().disableHtmlEscaping().create();
```
steven-solar, line 136: Worth looking into if this creates any security concerns. I doubt it will as all the data comes from the Google Books API, but disabling HTML escaping always seems a little risky.

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -125,41 +153,67 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+    // volumeInfo houses all of the information about the book itself: description,
+    // authors, title
+    JsonElement volumeInfo = bookInfo.get("volumeInfo");
+    if (volumeInfo == null)
+      return null;
+    JsonObject volumeInfoObj = volumeInfo.getAsJsonObject();
+
+    JsonElement titleElement = volumeInfoObj.get("title");
+    String title = titleElement != null ? titleElement.getAsString() : "";
```
steven-solar, line 189: Should this return null in the case that there is no title?

capstone-backend/src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -100,23 +147,45 @@ public void ensureCustomNumberOfBooksInResponse() {
...
+    VolumeData volume = SearchServlet.individualBookToVolumeData(bookJson);
+
+    Assert.assertTrue(checkPopulatedVolumeDataObject(volume));
+  }
+
+  @Test
+  public void checkErrorCodeWhenIncorrectParametersSent() {
+    String responseString = servletInvalidSearch();
+
+    Assert.assertTrue(responseString.isEmpty());
```
steven-solar, line 189: Very thorough testing!

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -45,29 +45,42 @@
...
+      if (request.getParameterMap().size() > 1) {
+        System.err.println("Error: No other parameter can be sent with a gbookId");
+        response.sendError(HttpServletResponse.SC_BAD_REQUEST);
+        return;
+      }
+      // make formatted url
+      String gbookId = request.getParameter("gbookId");
+      formattedURL = String.format("https://www.googleapis.com/books/v1/volumes/%s", gbookId);
+
+    } else if (request.getParameter("searchTerm") != null || !request.getParameter("searchTerm").isEmpty()) {
```
steven-solar, line 61: Should there also be a parameterMap size check inside this else-if clause? This would make sure nothing else is sent along with searchTerm.

steven-solar (approved):
Looks good!

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
antoniolinhart: Without disabling HtmlEscaping, strings like `\u003d` show up in place of other special characters like `=` or `&`. I think we can trust the Google Books API to return safe responses.

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
antoniolinhart: That's a good idea! If it doesn't have a non empty title, I don't think it should be returned since all books should at the very least have a title

MauroGuerrero20 (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
antoniolinhart: Yeah that's a good idea! I've now incorporated it using what you said when we discussed offline to check for 2 or more parameters, then confirm that maxResults is not null.

VivekShankar (approved):
LGTM - Do you know if there's a way to search for multiple gbookId's at once? I was just taking a look at Mauro's PR, specifically, the part where we fetch data for all the books on your book list by gbookId.

antoniolinhart:
> LGTM - Do you know if there's a way to search for multiple gbookId's at once? I was just taking a look at Mauro's PR, specifically, the part where we fetch data for all the books on your book list by gbookId.

@VivekShankar There is currently no way we know of to search for multiple gbookIds in the same query. It would be really nice if there was one!

PR #64, opened 2020-07-14
Change Assignments and Comments 

steven-solar:
In the last PR, a `Club` stored a list of `Assignments` and an `Assignment` stored a list of `Comments`. We realized now that rather than the "parent" holding a reference to the "child," it should be the other way around. If an assignment held a list of comments, then on comment creation we would need to perform two requests: a post to create the comment, and a put to update the list of comments in the assignment. The same issue arises for a club and its assignments. 

This PR alters this so that an Assignment holds its clubID, and a comment holds its assignmentID. The appropriate testing is updated.

### DIFF
--- a/capstone-backend/src/main/java/com/google/sps/data/Assignment.java
+++ b/capstone-backend/src/main/java/com/google/sps/data/Assignment.java
@@ -3,24 +3,18 @@ package com.google.sps.data;
 import java.util.ArrayList;
 import java.util.List;
 
+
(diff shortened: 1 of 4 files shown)


antoniolinhart (approved):
Looks good! This is a good change - this way it'll be much easier to search since the comments can be completely independent and not nested inside the object!

MauroGuerrero20 (approved):
LGTM

VivekShankar (approved):
LGTM

PR #69, opened 2020-07-15
BookSearchTile Test Updates 

### LINKED ISSUE #68: BookSearchTile Additional Tests
More tests need to be added to BookSearchTile to account for the dropdown tiles generated after the button is clicked. For now, the only tests that exist test the default view of the BookSearchTile without the dynamic changing after button click.

antoniolinhart:
# Description
After meeting with Geoff, I figured out how to test when a user 'clicks' on the Add to List button in the BookSearchTile and ensure that the correct dropdowns are displayed with the correct text!

Requires using act(), awaits and async because React needs to be able to make the updates asynchronously to the states and the other components.

# Linked Issue
#68


steven-solar (approved):
Looks great!

capstone-frontend/src/BookSearchTile.test.js
```diff
@@ -42,16 +42,22 @@ test('renders book search tile test', () => {
   const bookListButton = document.getElementById("dropdown-list-add");
   expect(bookListButton).toBeInTheDocument();
   expect(bookListButton.innerHTML).toEqual("Add to List");
-  
 
+  await act(async () => {bookListButton.click()});
+  const dropdownItems = document.getElementsByClassName("dropdown-item");
```
geoffbass: nit: Use single quotes for string literals (java/com/google/gsa/boq/shared/wishlist/proto/products.proto)

capstone-frontend/src/BookSearchTile.test.js
```diff
@@ -42,16 +42,22 @@ test('renders book search tile test', () => {
   const bookListButton = document.getElementById("dropdown-list-add");
   expect(bookListButton).toBeInTheDocument();
   expect(bookListButton.innerHTML).toEqual("Add to List");
-  
 
+  await act(async () => {bookListButton.click()});
```
geoffbass: I'm not familiar with `act`, but from the documentation it looks to me like you might not need to use `async`/`await` here, as long as you wrap any actions that have downstream UI effects you want to assert on in a call to `act`:
https://reactjs.org/docs/test-utils.html#act
https://reactjs.org/docs/testing-recipes.html#act

capstone-frontend/src/BookSearchTile.test.js
```diff
@@ -42,16 +42,22 @@ test('renders book search tile test', () => {
...
+test('renders book search tile with empty bookList test', async () => {
   render(<BookSearchTile book={book}
-    userBookLists={emptyArray} />, container);
+    bookLists={emptyArray} />, container);
 
   const bookListButton = document.getElementById("dropdown-list-add");
   expect(bookListButton).toBeInTheDocument();
   expect(bookListButton.innerHTML).toEqual("No Lists Found");
 
+  await act(async () => {bookListButton.click()});
```
geoffbass: Same here: see if it still works when omitting `async`/`await`

capstone-frontend/src/BookSearchTile.test.js
```diff
@@ -42,16 +42,22 @@ test('renders book search tile test', () => {
   const bookListButton = document.getElementById("dropdown-list-add");
   expect(bookListButton).toBeInTheDocument();
   expect(bookListButton.innerHTML).toEqual("Add to List");
-  
 
+  await act(async () => {bookListButton.click()});
+  const dropdownItems = document.getElementsByClassName("dropdown-item");
+  expect(dropdownItems.length).toEqual(3);
+  expect(dropdownItems[0].innerHTML).toEqual("<span>Best Books</span>");
```
geoffbass: The fact that the text is wrapped in a `span` is really an implementation detail that would be best not to test directly. See if there's a way you can assert on only the text content here.

geoffbass (approved):
LGTM, good work finding the `act` utility function!

MauroGuerrero20 (approved):
LGTM 

Remmber to sync with master. Some of my merged PR may could result in the test failing.

capstone-frontend/src/BookSearchTile.test.js
antoniolinhart: Fixed!

capstone-frontend/src/BookSearchTile.test.js
antoniolinhart: This second case was removed since the other conditional button was made a regular button instead of a dropdown button requiring a click to 'activate'

capstone-frontend/src/BookSearchTile.test.js
antoniolinhart: I changed this to `toMatch(/Create New List/)` with the innerHTML, because I try to get .innerText it gives me nothing. In this way, I don't care about the span, just about the Create New List text.

capstone-frontend/src/BookSearchTile.test.js
antoniolinhart: Discussed offline and couldn't find a way to avoid the use of async and await here.

antoniolinhart:
> LGTM
> 
> Remmber to sync with master. Some of my merged PR may could result in the test failing.

Merged with master and resolved conflicts in the testing file!

PR #82, opened 2020-07-18
Club Page: UserCards, Admin Page, Styling

steven-solar:
This PR makes changes to the `ClubPage` and creates the `ClubAdminPage`. 

* `UserCard` to display user's names and profile pictures. In the case that they are used in a Club, and the current logged in user is the club's owner, a Remove Member button is rendered. Clicking this will remove the user from the club.
* `AdminClubPage` allows the club's owner to change the club's name and description, and delete the club. A button to navigate to this page is rendered conditionally if the current logged in user is the club's owner. The requests performed on this page are reliant on the logged in user being the club's owner. 
* Commenting now triggers a re-render of the `AssignmentCard`. 
* Assignments can be posted on the ClubPage if the logged in user is the club owner
* React Bootstrap is used to standardize form and button styles, later PRs will refactor the `CreateClub` component to also use the library's components

See admin toggle button at top, as I am signed in as the club owner, as well as the owner's `UserCard` at the top of the page (no Remove Member button): 
Screenshot 2020-07-20 at 10 15 57 AM

See commenting altered to use React Bootstrap forms, re-renders an assignment card on comment:
Screenshot 2020-07-20 at 10 18 22 AM

See the post assignment form (only visible because I am signed in as the club's owner), and the UserCards of its members (with Remove Member buttons since I am the club owner): 
Screenshot 2020-07-20 at 10 18 27 AM

See the `AdminClubPage`, which is only reachable since I am the club owner (future PRs will expand the functionality of this page):
Screenshot 2020-07-20 at 10 18 33 AM


capstone-frontend/src/components/AdminClubPage.jsx
```diff
@@ -0,0 +1,93 @@
...
+    const history = this.props.history;
+    e.preventDefault();
+    let data = {};
+    const formElements = document.getElementById("update-club-form").elements;
+    for (let i = 0; i < formElements.length; i++) {
+      if (formElements[i].name.length !== 0 && 
+          formElements[i].type !== "submit" && formElements[i].value.length !== 0) {
+        data[formElements[i].name] = formElements[i].value;
+      }
+    }
```
antoniolinhart, line 34: This seems like a good way to adapt to a growing list of update fields!

capstone-frontend/src/components/AdminClubPage.jsx
```diff
@@ -0,0 +1,93 @@
...
+            alert(err);
+        });
+  }
+
+  componentDidMount() {
+    this.fetchClub();
+  }
+
+  render() {
+    return (
```
antoniolinhart, line 72: I really like the look of this admin page!

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -57,16 +70,23 @@ class ClubPage extends Component {
...
-    fetch(`/api/clubs?id=${this.props.match.params.id}`, {method: "delete"})
-        .then(function() {
-            history.push("/myclubs");
-        })
+
+  handleAssignmentPost = (e) => {
+    e.preventDefault();
+    if (window.localStorage.getItem("userID") !== this.state.club.ownerID) {
+      alert("Assignment creation failed. You do not own this club.");
+      return;
```
antoniolinhart, line 78: Even though you shouldn't be able to see this button without being an owner, I like this check to ensure that nonowners can't post assignments since they would have to somehow fake their userId

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -75,19 +95,40 @@ class ClubPage extends Component {
...
-        {members}
+        <div> Club Owner: </div>
+        <Row className="align-items-center justify-content-center">
+          {owner}
+        </Row>
         <div className="description"> {this.state.club.description} </div>
         {bookTile}
         {assignments}
-        <button onClick={this.handleClick}> Delete </button>
+        {isOwner &&
```
antoniolinhart, line 125: This is a cool toggle to ensure that it should only get displayed if they are an owner!

capstone-frontend/src/components/CommentCard.jsx
```diff
@@ -30,7 +30,7 @@ class CommentCard extends Component {
       <div className="comment-border">
         {user}
         <div> {this.props.comment.text} </div>
-        {/* <div> {new Date(this.props.comment.whenCreated)} </div> */}
+        <div> {(new Date(this.props.comment.whenCreated)).toString()} </div>
```
antoniolinhart: Does this .toString() resolve the minified React errors we were having last week?

capstone-frontend/src/components/UserCard.jsx
```diff
@@ -0,0 +1,38 @@
...
+                             && this.props.club.ownerID === window.localStorage.getItem("userID")
+                             && this.props.club.ownerID !== this.props.user.id
+                             && <Button id="remove-member" variant="danger" onClick={this.removeMember}>
+                                  Remove&nbsp;Member
+                                </Button>;
+    return (
+      <Col className="user-card" xs={{span: "2"}} >
+        <img id="user-profile" src={this.props.user.profileImageUrl} alt="Profile"/>
+        <div> {this.props.user.fullName} </div> 
+        <Button variant="primary"> Add Friend </Button>
```
antoniolinhart: I'd recommend putting a function call for the onClick of this button and making a stub with an issue number so we can more easily track it and implement it!

capstone-frontend/src/styles/Groups.css
```diff
@@ -63,3 +63,32 @@ form div {
...
+  border-radius: 15px;
+  box-shadow: 3px 3px 3px #888888;
+  padding: 15px 10px;
+  margin: 10px;
+}
+
+#user-profile {
+  border-radius: 50%;
+}
+
```
antoniolinhart: This part of the CSS is alphabetized, you should also alphabetize it in the rest of the styles in this file.

antoniolinhart (approved):
LGTM

capstone-frontend/src/components/AdminClubPage.jsx
steven-solar: Thanks!

capstone-frontend/src/components/AdminClubPage.jsx
steven-solar: Thank you! React bootstrap makes styling super easy and consistent across components.

capstone-frontend/src/components/ClubPage.jsx
steven-solar: Thanks!

capstone-frontend/src/components/CommentCard.jsx
steven-solar: It does. I also opened a separate PR for just the fix, just so it was easier to merge or we want to roll back this PR for whatever reason.

capstone-frontend/src/components/UserCard.jsx
steven-solar: Good idea. I'll add that.

capstone-frontend/src/styles/Groups.css
steven-solar: I'll fix that!

MauroGuerrero20 (approved):
LGTM

https://github.com/STEP-Pod-Capstone-Project/Capstone-Project/pull/83 Should probably be merge first.

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -57,16 +70,23 @@ class ClubPage extends Component {
...
+      alert("Assignment creation failed. You do not own this club.");
+      return;
+    }
+    let data = {
+      "clubID": this.state.club.id,
+      "text": e.target[0].value,
+      "whenCreated": (new Date()).toUTCString()
+    };
+    fetch(`/api/assignments`, {method: "post", body: JSON.stringify(data)})
+        .then(this.fetchData())
```
VivekShankar: Is it necessary to call `fetchData` when an assignment is posted? Wondering if you could get away with just making a call to the assignments servlet here to update the club page after a new assignment is posted.

capstone-frontend/src/components/UserCard.jsx
```diff
@@ -0,0 +1,38 @@
...
+        method: "put", 
+        body: JSON.stringify(jsonBody)
+      })
+      .then(this.props.removeMember(this.props.user.id))
+  }
+
+  render() {
+    const removeMember = this.props.club 
+                             && this.props.club.ownerID === window.localStorage.getItem("userID")
+                             && this.props.club.ownerID !== this.props.user.id
```
VivekShankar, line 27: Nice checks for potential edge cases here! Just for clarification, are you ensuring that only the club owner can remove members, and additionally, that the club owner can't remove themselves from the club?

VivekShankar (approved):
LGTM

capstone-frontend/src/components/UserCard.jsx
steven-solar: Thanks! Yes, the idea is that only the club owner can remove members. Additionally, the club owner should not be able to be removed from the club (even though the only person who can do this is themselves).

capstone-frontend/src/components/ClubPage.jsx
steven-solar: What I could do is break `fetchData` into several helper methods (ie. one for the club, one for assignments, one for the book, etc) this way I can just call `fetchAssignmentData` after an assignment post and re-fetch just the assignments rather than all the data for the whole page. 

I think an even better solution is to take the response json (which will be the created assignment) and append it to the state's assignments without re-fetching to trigger a re-render. This will be much quicker than refetching all the club data on a post, and was the reason I originally had the post method return the created object. I'll go with this option.

steven-solar: I'm going to implement this same logic in my `onComment` function, as I was previously re-fetching all an assignment's comments onComment. Now I'll use the response from the post to update the state appropriately and reduce unnecessary fetches.

PR #83, opened 2020-07-19
Fix date bug

steven-solar:
This PR fixes the previous bug with rendering `Date` objects in React. This is done by calling the `.toString()` method on the Dates when placing them inside components.


antoniolinhart (approved):
LGTM. Do you know why toString fixes the problem?

steven-solar:
> LGTM. Do you know why toString fixes the problem?

The problem was that we were rendering an `Object` directly inside a React component, which React did not like. When I first began using the dates I assumed they would by default call their `toString` when placed inside components (like they would inside a `console.log`). However, this is not the case and we need to explicitly call `toString()` so they become strings and not JS objects.

MauroGuerrero20 (approved):
LGTM

VivekShankar (approved):
LGTM

Issue #72, opened 2020-07-16
SearchServlet Thumbnail Retrieval


antoniolinhart:
When querying the Google Books API for a thumbnail, set the variable to be the link of the largest thumbnail as not all sizes are available for every book.

Thumbnail order: `extraLarge`, `large`, `medium`, `small`, `thumbnail`

PR #87, opened 2020-07-20
Book Thumbnail Link

antoniolinhart:
# Description

This PR focuses on the implementation of getting the largest thumbnail link available from the Google Books API call. This order is `extraLarge`, `large`, `medium`, `small`, `thumbnail`. Every book is guaranteed to at least have a `thumbnail` element, but it might not have any sizes of greater quality.

From what I can tell, the books retrieved using the `query` (so, the search for a book using a query parameter) only come with a `thumbnail` element and an equivalent or smaller `smallThumbnail`. Not all books retrieved through individual retrieval have the larger thumbnail sizes, but it is better to display the largest image we can to not show blurry images on the frontend.

# Linked Issue
#72

### DIFF
--- a/capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
@@ -29,6 +29,7 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.URLEncoder;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.stream.Collectors;
 import javax.servlet.annotation.WebServlet;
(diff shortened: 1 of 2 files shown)


MauroGuerrero20 (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -212,8 +213,14 @@ static VolumeData individualBookToVolumeData(JsonElement singleBook) {
     String canonicalVolumeLink = canonVolumeElement != null ? canonVolumeElement.getAsString() : "";
 
     JsonElement thumbnailElement = volumeInfoObj.get("imageLinks");
-    String thumbnailLink = thumbnailElement != null ? thumbnailElement.getAsJsonObject().get("thumbnail").getAsString()
-        : "";
+    ArrayList<String> sizes = new ArrayList<>(Arrays.asList("extraLarge", "large", "medium", "small", "thumbnail"));
```
steven-solar, line 217: Why is the last size "thumbnail"?

steven-solar (approved):
Looks great! A really good solution to use the best images possible on the book page you created.

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -212,8 +213,14 @@ static VolumeData individualBookToVolumeData(JsonElement singleBook) {
...
+
+    String thumbnailLink = "";
+    String greatestSize = sizes.stream().filter(size -> thumbnailElement.getAsJsonObject().get(size) != null)
```
VivekShankar: Nice use of streams and `findFirst` here!

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -212,8 +213,14 @@ static VolumeData individualBookToVolumeData(JsonElement singleBook) {
...
+        .findFirst().get();
```
VivekShankar: Note that calling `Optional.get` here will throw a `NoSuchElementException` if none of the sizes make it past the previous `filter` step. The fact that the `"thumbnail"` element is guaranteed to be there should ensure this won't happen, but I think it might be a good idea to program defensively here. For example, instead of calling `.get` here, you could instead add an `ifPresent` check to the conditional below.

VivekShankar (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
antoniolinhart: `thumbnail` is the largest default image given by the API, you can see in this example here: https://www.googleapis.com/books/v1/volumes/x0BHx72wj1sC

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
antoniolinhart: Fixed! I didn't know about the Optional data type in Java but that is interesting!

PR #76, opened 2020-07-16
Book Page

antoniolinhart:
# Description

This PR focuses on the development of the Book Page which is a page showing the title of the book, its description, rating, and links to a preview and the Google Play page for the book.

There will be another PR that implements the routing to this BookPage with the book's gbookId when a book is clicked on.

# Breakdown
- Uses SearchServlet API to query for the book's information based on its gbookId
- Extract BookListAddDropdown into another class so it can be used here on this page as well
- Uses React-Star-Ratings package (https://www.npmjs.com/package/react-star-ratings) to easily show the average rating of the book on the site
  - If there are no ratings, 0/5 stars will be shown and it will include a "no ratings" message
- Abstracted some CSS code into a shared "Book.css" class so it is not redefined in two places

# Main Difficulties

Figuring out how button sizing works to ensure both buttons on the page were the same size.

Setting up bootstrap containers and rows/columns to ensure the proper layout of the components

Updating code to reflect other changes to the BookSearchTile

# Screenshot of Feature

BookPage image

BookPage mobile image

# Linked Issue
#77

PR #92, opened 2020-07-21
Book Page Link Integration

### LINKED ISSUE #79: Integrate BookPage with BookListPage
- When the user clicks an image of a book while on the ```ListPage``` it should redirect them to @antoniolinhart ```BookPage```

antoniolinhart:
# Description

This PR focuses on the integration of the BookPage link in the other pages of the application: `BookSearchTile`, `CreateList`, `ListPage` and `SearchBookModal`. 

It creates the BookDescriptionOverlay, a component used to wrap around the book image and give a description popup when the user clicks on the image.

Finishes work started in #76 

# Breakdown
- Create BookDescriptionOverlay using react bootstrap popovers
  - Renders all children inside of it
  - If there is a description over 500 chars, cut off and add ellipsis
  - If there is an empty description, do not show the description tag
- Add BookDescriptionOverlay to `BookSearchTile`, `CreateList`, `ListPage` and `SearchBookModal`
- Add image classes in the CSS to ensure the sizes are as expected and not too large
- Minor string literal changes

# Main Difficulties

Deciding whether to implement this as a function in each necessary class or make a new component for it.

Placing the Overlay simply on the modals, SearchBookModal and CreateList, or all pages referencing books.

# Screenshot of Feature

This is the description overlay shown when clicking on a book image.

Populated Description Overlay

This is the description overlay shown when clicking on a book image with no description.

Empty Description Overlay

# Linked Issue
#79

### DIFF
--- a/capstone-frontend/src/components/BookDescriptionOverlay.jsx
+++ b/capstone-frontend/src/components/BookDescriptionOverlay.jsx
@@ -1,11 +1,10 @@
 import React from 'react';
+import { Link } from 'react-router-dom';
 import { Popover, OverlayTrigger } from 'react-bootstrap';
 
 const popoutDescription = (book) => {
-  let description = book.description;
-  if (description === undefined) {
-    description = "";
-  }
+  let description = book.description || "";
+
   if (description.length > 500) {
     description = description.substring(0, 500).concat('...');
   }
@@ -13,19 +12,18 @@ const popoutDescription = (book) => {
   return (
     <Popover>
       <Popover.Title as='h3'>
-        <a className='center-horizontal'
-          href={`/bookpage/${book.id}`} target='_blank'
-          rel='noopener noreferrer'>
+        <Link to={`/bookpage/${book.id}`} className='center-horizontal'
+          target='_blank' rel='noopener noreferrer'>
           {book.title}
-        </a>
+        </Link>
       </Popover.Title>
-      {description.length > 0 &&
+      {description &&
         <Popover.Content>
           <p className='font-weight-bold'>Description:</p>
           {description}
         </Popover.Content>
       }
-    </Popover >
+    </Popover>
   );
 }
 
--- a/capstone-frontend/src/components/BookDescriptionOverlay.jsx
+++ b/capstone-frontend/src/components/BookDescriptionOverlay.jsx
@@ -26,7 +26,7 @@ const popoutDescription = (book) => {
 
 const BookDescriptionOverlay = (props) => {
   return (
-    <OverlayTrigger trigger="hover" placement="right" overlay={popoutDescription(props.book)}>
+    <OverlayTrigger trigger='hover' placement='right' overlay={popoutDescription(props.book)}>
       {props.children}
     </OverlayTrigger>
   );
--- a/capstone-frontend/src/components/BookDescriptionOverlay.jsx
+++ b/capstone-frontend/src/components/BookDescriptionOverlay.jsx
@@ -26,7 +26,7 @@ const popoutDescription = (book) => {
 
 const BookDescriptionOverlay = (props) => {
   return (
-    <OverlayTrigger trigger='hover' placement='right' overlay={popoutDescription(props.book)}>
+    <OverlayTrigger trigger='click' placement='right' overlay={popoutDescription(props.book)}>
       {props.children}
     </OverlayTrigger>
   );
--- a/capstone-frontend/src/components/BookDescriptionOverlay.jsx
+++ b/capstone-frontend/src/components/BookDescriptionOverlay.jsx
@@ -1,0 +1,35 @@
+import React from 'react';
+import { Popover, OverlayTrigger } from 'react-bootstrap';
+
+const popoutDescription = (book) => {
+  let description = book.description;
+  if (description.length > 500) {
+    description = description.substring(0, 500).concat('...');
+  }
+
+  return (
+    <Popover>
+      <Popover.Title as='h3'>
+        <a className='center-horizontal'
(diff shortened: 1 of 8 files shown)


capstone-frontend/src/components/BookDescriptionOverlay.jsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { Popover, OverlayTrigger } from 'react-bootstrap';
+
+const popoutDescription = (book) => {
+  let description = book.description;
```
steven-solar: You could combine the assignment in lines 5-8 into one line using javascript short-circuit expressions. Check this out for guidance: https://codeburst.io/javascript-what-is-short-circuit-evaluation-ff22b2f5608c. The following assignment:
```
let description = book.description || "";
``` 
would have the same behavior as your current implementation and be a little cleaner.

capstone-frontend/src/components/BookDescriptionOverlay.jsx
```diff
@@ -0,0 +1,40 @@
...
+    description = "";
+  }
+  if (description.length > 500) {
+    description = description.substring(0, 500).concat('...');
+  }
+
+  return (
+    <Popover>
+      <Popover.Title as='h3'>
+        <a className='center-horizontal'
```
steven-solar: I think we should use a react-router `Link` here instead of an anchor tag. In `App.jsx` we define the `Route` for "/bookpage/:id" and I think it would be cleaner to pair that Route with a corresponding Link here.

capstone-frontend/src/components/BookDescriptionOverlay.jsx
```diff
@@ -0,0 +1,40 @@
...
+          href={`/bookpage/${book.id}`} target='_blank'
+          rel='noopener noreferrer'>
+          {book.title}
+        </a>
+      </Popover.Title>
+      {description.length > 0 &&
```
steven-solar: Seems like you could let `description` be `undefined` earlier and change this expression to take that into account (ie change to `description && ...`), rather than making it an empty string, only to not display it.

steven-solar (approved):
This looks great! I left a few small comments, but I really like this addition.

MauroGuerrero20 (approved):
LGTM

capstone-frontend/src/components/BookDescriptionOverlay.jsx
antoniolinhart: I like this change!

capstone-frontend/src/components/BookDescriptionOverlay.jsx
antoniolinhart: This is a good idea! I knew about the && but not using || in this case

capstone-frontend/src/components/BookDescriptionOverlay.jsx
antoniolinhart: Added!

capstone-frontend/src/components/BookDescriptionOverlay.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import { Popover, OverlayTrigger } from 'react-bootstrap';
+
+const popoutDescription = (book) => {
```
geoffbass, line 8: This looks like it should really be treated as a stateless functional component, since it returns JSX.

capstone-frontend/src/components/BookDescriptionOverlay.jsx
```diff
@@ -0,0 +1,38 @@
...
+  let description = book.description || "";
```
geoffbass, line 9: nit: Use single quotes for string literals (https://google.github.io/styleguide/jsguide.html#features-strings-use-single-quotes).

capstone-frontend/src/components/BookDescriptionOverlay.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+  if (description.length > 500) {
+    description = description.substring(0, 500).concat('...');
+  }
+
+  return (
+    <Popover>
+      <Popover.Title as='h3'>
+        <Link to={`/bookpage/${book.id}`} className='center-horizontal'
```
geoffbass, line 18: nit: This would probably be more readable with each prop on its own line. Follow this pattern when they don't all fit on the same line.

capstone-frontend/src/components/CreateList.jsx
```diff
@@ -115,7 +116,7 @@ class CreateList extends Component {
 
   render() {
     return (
-      <div>
+      <>
```
geoffbass, line 119: Nice use of fragments!

geoffbass (approved):
LGTM

capstone-frontend/src/components/BookDescriptionOverlay.jsx
antoniolinhart: True, this does make it much more readable and consistent!

capstone-frontend/src/components/BookDescriptionOverlay.jsx
antoniolinhart: Fixed!

capstone-frontend/src/components/BookDescriptionOverlay.jsx
antoniolinhart: Treating this as a stateless functional component caused the popout to render at the top left of the screen instead of next to the book the user was clicking on. 

Discussed this offline and was not able to figure out a way to use the PopoutDescription component in the overlay call and produce the desired result. Left a comment above the function explaining the reasoning.

PR #97, opened 2020-07-22
SearchBookModal in ClubPage

steven-solar:
This PR integrates the `SearchBookModal` into the `ClubPage`. With a few small tweaks to the Modal, it can now be used to set the book a club is reading. The modal is only displayed if the current user is the club's owner. 

Screenshot 2020-07-22 at 11 31 02 AM
Screenshot 2020-07-22 at 11 31 13 AM
Screenshot 2020-07-22 at 11 35 42 AM

### DIFF
--- a/capstone-frontend/src/components/ClubPage.jsx
+++ b/capstone-frontend/src/components/ClubPage.jsx
@@ -74,16 +74,16 @@ class ClubPage extends Component {
 
   handleAssignmentPost = (e) => {
     e.preventDefault();
-    if (window.localStorage.getItem("userID") !== this.state.club.ownerID) {
-      alert("Assignment creation failed. You do not own this club.");
+    if (window.localStorage.getItem('userID') !== this.state.club.ownerID) {
+      alert('Assignment creation failed. You do not own this club.');
       return;
     }
     let data = {
-      "clubID": this.state.club.id,
-      "text": e.target[0].value,
-      "whenCreated": (new Date()).toUTCString()
+      'clubID': this.state.club.id,
+      'text': e.target[0].value,
+      'whenCreated': (new Date()).toUTCString()
     };
-    fetch(`/api/assignments`, {method: "post", body: JSON.stringify(data)})
+    fetch(`/api/assignments`, {method: 'post', body: JSON.stringify(data)})
         .then(response => response.json())
         .then(assignmentJson => {
           let assignments = this.state.assignments;
@@ -97,7 +97,9 @@ class ClubPage extends Component {
   }
 
   handleBookChange = async (gbookID) => {
-    this.state.club.gbookID = gbookID;
+    let club = this.state.club;
+    club.gbookID = gbookID;
+    this.setState({club});
     await fetch(`/api/search?gbookId=${this.state.club.gbookID}`)
         .then(response => response.json()).then(bookJson => this.setState({book: bookJson[0]}))
         .catch(function(err) {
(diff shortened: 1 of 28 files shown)


capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -95,6 +96,16 @@ class ClubPage extends Component {
         });
   }
 
+  handleBookChange = async (gbookID) => {
+    this.state.club.gbookID = gbookID;
```
antoniolinhart: You should use setState instead of setting state like this since setState will trigger a react rerender.

capstone-frontend/src/components/SearchBookModal.jsx
```diff
@@ -58,21 +58,11 @@ export class SearchBookModal extends Component {
...
-          addedBooksIDs: [book.id],
-          addedBooks: [book]
-        }
-      )
-    }
-    else {
-      this.state.addedBooksIDs.push(book.id);
-      this.state.addedBooks.push(book);
+    this.state.addedBooksIDs.push(book.id);
+    this.state.addedBooks.push(book);
```
antoniolinhart: In this case, instead of having this with the rerender below I think you can do something like this:
`this.setState({ addedBookIDs: [...this.state.addedBookIDs, book.id] })`

Taken from https://stackoverflow.com/questions/37435334/correct-way-to-push-into-state-array

capstone-frontend/src/components/SearchBookModal.jsx
```diff
@@ -128,6 +109,19 @@ export class SearchBookModal extends Component {
...
 
+  clubSubmit = async (gbookID) => {
+    let updateJson = {
+      id: this.props.objectId,
+      gbookID: gbookID,
+    }
+    await fetch(this.props.putURL, {
+      method: "PUT",
+      body: JSON.stringify(updateJson)
+    });
```
antoniolinhart: Should we have a catch on this to handle any errors if the PUT doesn't work?

antoniolinhart (approved):
I like how this component can be adapted for multiple uses!

LGTM

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -121,6 +132,15 @@ class ClubPage extends Component {
         </Row>
         <div className="description"> {this.state.club.description} </div>
         {bookTile}
+        {isOwner && 
+          <SearchBookModal
+            objectId={this.props.id}
+            update={this.handleBookChange}
+            text="Change the Club's Book"
+            putURL="/api/clubs"
```
geoffbass: nit: Use single quotes for string literals (https://google.github.io/styleguide/jsguide.html#features-strings-use-single-quotes).

capstone-frontend/src/components/SearchBookModal.jsx
```diff
@@ -95,19 +85,10 @@ export class SearchBookModal extends Component {
...
-            id: this.props.objectId,
-            gbookID: bookId,
-          }
-        }
-        else {
-          updateJson = {
-            id: this.props.objectId,
-            add_gbookIDs: bookId,
-          }
+        updateJson = {
```
geoffbass, line 84: nit: This will be turned into JSON in the request body but is just a regular object right now, so naming it something like "updates" would be more accurate.

capstone-frontend/src/components/SearchBookModal.jsx
```diff
@@ -128,6 +109,19 @@ export class SearchBookModal extends Component {
     }
   }
 
+  clubSubmit = async (gbookID) => {
+    let updateJson = {
```
geoffbass: nit: `const` (https://google.github.io/styleguide/jsguide.html#features-use-const-and-let)

capstone-frontend/src/components/SearchBookModal.jsx
```diff
@@ -179,7 +173,11 @@ export class SearchBookModal extends Component {
                         {this.state.addedBooksIDs.includes(book.id) ?
                           <Button className="my-5" variant="danger" onClick={() => this.removeBookFromList(book)}>Remove Book</Button>
                           :
-                          <Button className="my-5" onClick={() => this.addBookToList(book)}>Add to Booklist</Button>}
+                          this.props.type === "club" ?
+                            <Button className="my-5" onClick={async () => this.clubSubmit(book.id)}>Add</Button>
```
geoffbass: nit: Don't think this needs to be `async`

capstone-frontend/src/components/SearchBookModal.jsx
```diff
@@ -222,11 +220,13 @@ export class SearchBookModal extends Component {
                 )
               }
 
-              <div className="text-center">
-                <Button className="text-center" variant="primary" onClick={async () => await this.handleSubmit()} >
-                  Add
+              {this.props.type !== "club"
+                && <div className="text-center">
+                  <Button className="text-center" variant="primary" onClick={async () => await this.handleSubmit()} >
```
geoffbass: Same here: pretty sure the `async` and `await` are unnecessary here

geoffbass (approved):
LGTM

capstone-frontend/src/components/ClubPage.jsx
steven-solar: I'll change that. Only reason I didn't was I was calling setState below anyways, but can't hurt to call in both places.

capstone-frontend/src/components/SearchBookModal.jsx
steven-solar: I can change this, but this function is solely used for the BookList implementation and didn't want to play with anything not related to my feature in case it alters how Mauro uses it. I can discuss with him

capstone-frontend/src/components/SearchBookModal.jsx
steven-solar: Good call.

capstone-frontend/src/components/ClubPage.jsx
steven-solar: I'll change this!

MauroGuerrero20 (approved):
LGTM

capstone-frontend/src/components/SearchBookModal.jsx
steven-solar: Will fix

capstone-frontend/src/components/SearchBookModal.jsx
steven-solar: I'll fix this.

capstone-frontend/src/components/SearchBookModal.jsx
steven-solar: This part is only used for Mauro's BookLists so I'll check with him before changing anything, but I think you're right.

steven-solar: Spoke with Mauro, we changed and it works fine!

capstone-frontend/src/components/SearchBookModal.jsx
steven-solar: Spoke with Mauro. We made the change and it works fine! Updating and merging.

PR #106, opened 2020-07-24
Updated Browse, Club Search Page

antoniolinhart:
# Description

This PR focuses on implementing a way to search for clubs on the Browse tab. Originally, the user could only enter a search term and get a list of books matching that search term, but now the user can get a list of clubs based on the search term as well.

A new ClubSearchList was added to perform the search for Clubs by using a new ClubSearchServlet with prefix matching based on the name of the clubs.

# Breakdown
- Create ClubSearchServlet to perform prefix string matching on the club names
  - Gets a list of all clubs matching the search query which is called when the user types in a search term
- Added menu to toggle between search for Books, Clubs, and Users
  - Users is currently just a stub
- Made browse a stateful component
- Added headers so the user knows what they are searching for: searching for Books with "searchTerm"
- Moved retrieval of searchedBooks to the Browse component since it is being used by ClubSearchList and BookSearchList
  - Made BookSearchList a stateless functional component, it just takes in a list of Book items to display now
- ClubSearchList has 2 different searches
  - Prefix matching of names of clubs
  - Takes in books from searchedBooks and checks if any clubs contain those books that were fetched
- Minor fixes
 - Added loading spinners for the Browse page while it is waiting for the data to be returned
 - Added img-fluid on BookPage and BookSearchTile to prevent overlapping elements due to thumbnail size

# Main Difficulties

I decided to create a new ClubSearchServlet instead of adding an if-check to the original ClubServlet because I did not want to make it too cluttered. This ClubSearch is not likely to be used extensively across the application, while the doGet inside the ClubServlet is used in many places.

I decided to make Browse a stateful component since it needed to fetch and store the books that would be required by the BookSearchList and ClubSearchList.

# Screenshot of Feature

This is the new default search view when someone enters a new search term into the box.

New Book Search View With Tabs

This is the new club search view when the user clicks on the Clubs tab!

New Club Search View With Tabs

Mobile view updated to work with the new tabs.

Mobile Book Search View With Tabs


capstone-backend/src/main/java/com/google/sps/servlets/ClubSearchServlet.java
```diff
@@ -0,0 +1,58 @@
+package com.google.sps.servlets;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Map;
```
MauroGuerrero20: Map is never used.

capstone-frontend/src/components/Browse.jsx
```diff
@@ -1,10 +1,103 @@
...
+import { Nav, Row, Spinner, Tab } from 'react-bootstrap';
+import { BookSearchList } from './BookSearchList';
+import { ClubSearchList } from './ClubSearchList';
 
-const Browse = (props) => {
-  return (
-    <BookSearchList searchQuery={props.searchQuery} bookLists={props.bookLists} updateBookLists={props.updateBookLists} />
-  );
-}
+import '../styles/Browse.css';
```
MauroGuerrero20: Browse.css is empty. I would delete it if it servers no purpose.

MauroGuerrero20 (approved):
LGTM

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -9,19 +9,19 @@ const BookSearchTile = (props) => {
     <div className='book-search-tile'>
       <Container>
         <Row className='justify-content-md-center'>
-          <Col md='auto'>
+          <Col md={3}>
```
steven-solar: Should this be different based on different screen sizes (ie. xs=12, sm=6, md=3)?

steven-solar (approved):

capstone-backend/src/main/java/com/google/sps/servlets/ClubSearchServlet.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+import com.google.cloud.firestore.CollectionReference;
+import com.google.cloud.firestore.DocumentSnapshot;
+import com.google.cloud.firestore.Firestore;
+import com.google.cloud.firestore.Query;
+import com.google.gson.Gson;
+import com.google.sps.data.Club;
+
+@WebServlet("/api/clubSearch")
```
geoffbass: The existing 'clubs' servlet actually looks pretty lean at the moment. I think it might make more sense to make this a part of that, or to turn this into a sort of 'sub'-servlet if the framework allows that sort of thing.

capstone-backend/src/main/java/com/google/sps/servlets/ClubSearchServlet.java
```diff
@@ -0,0 +1,58 @@
...
+  public ClubSearchServlet() throws IOException {
+    db = Utility.getFirestoreDb();
+    clubs = db.collection("clubs");
+    gson = new Gson();
+  }
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    if (request.getParameter("searchTerm") == null || request.getParameterMap().size() > 1) {
+      System.err.println("Error: Improper parameters, must only send searchTerm");
```
geoffbass: It might be useful to split these out with separate error messages for more targeted debugging.

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -46,7 +46,6 @@
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-
```
geoffbass: nit: Should probably be in a separate formatting PR. Have you looked into setting up an auto-formatter?

capstone-frontend/src/components/Browse.jsx
```diff
@@ -1,10 +1,103 @@
...
+    this.state = {
+      searchedBooks: [],
+      loading: true, // Spinner
+    }
+  }
+
+  getSearchedBooks = (searchQuery) => {
+    fetch(`/api/search?searchTerm=${searchQuery}`)
+      .then(response => response.json())
+      .then(searchedBooks => this._isMounted && this.setState({ searchedBooks, loading: false }))
```
geoffbass, line 18: This looks like it should work, but I would spend a little time trying to figure out why the component is un-mounting with an outstanding network request if you haven't already.

capstone-frontend/src/components/Browse.jsx
```diff
@@ -1,10 +1,103 @@
...
+      .catch(function (err) {
+        alert(err)
+      });
+  }
+
+  componentDidMount() {
+    this._isMounted = true;
+
+    if (this._isMounted) {
```
geoffbass: This should always be `true` given line 27 above

capstone-frontend/src/components/Browse.jsx
```diff
@@ -1,10 +1,103 @@
...
+    }
+  }
+
+  componentWillUnmount() {
+    this._isMounted = false;
+  }
+
+  render() {
+    return (
+      <>
```
geoffbass: I don't think the fragment is necessary here if everything is wrapped in the `Tab.Container`

capstone-frontend/src/components/ClubSearchList.jsx
```diff
@@ -0,0 +1,104 @@
...
+    // Get clubs by searchQuery
+    let clubs = await fetch(`/api/clubSearch?searchTerm=${this.props.searchQuery}`)
+      .then(response => response.json())
+      .catch(function (err) {
+        alert(err)
+      });
+
+    // Get clubs by checking to see if any clubs are reading the books found in Google Books API search
+    await Promise.all(this.props.books.map(async (book) => {
+      const clubsWithBook = await fetch(`/api/clubs?gbookID=${book.id}`)
```
geoffbass: The `await` here makes it so that each network request only goes out once the previous one has completed the round trip. You should be able to omit `await` and modify the `clubs` array in your `.then` callback to do the processing in parallel, same for the requests generated inside loops below. In general, `await` is only needed when the subsequent steps need access to the returned data, otherwise you're unnecessarily throttling yourself.

capstone-frontend/src/components/ClubSearchList.jsx
```diff
@@ -0,0 +1,104 @@
...
+      }
+      if (club.ownerID.length > 0) {
+        const owner = await fetch(`/api/user?id=${club.ownerID}`)
+          .then(response => response.json())
+          .catch(function (err) {
+            alert(err)
+          });
+        club.ownerName = owner.fullName;
+      }
+    }));
```
geoffbass: The logic here looks good to me but I would prefer to see as much of it moved to the back-end as possible

capstone-frontend/src/components/ClubSearchList.jsx
```diff
@@ -0,0 +1,104 @@
...
+
+    this._isMounted && this.setState({ clubs, loading: false });
+  }
+
+  componentDidMount() {
+    this._isMounted = true;
+
+    if (this._isMounted) {
```
geoffbass: Same here: should always be `true`

capstone-frontend/src/components/ClubSearchList.jsx
```diff
@@ -0,0 +1,104 @@
...
+      this.setState({ loading: true });
+      this.getData();
+    }
+  }
+
+  componentDidUpdate(prevProps) {
+    if (this.props.searchQuery !== prevProps.searchQuery
+      || this.props.books !== prevProps.books) {
```
geoffbass: nit: I like splitting the condition into two lines but the second line should be indented to follow the [rectangle rule](https://github.com/google/google-java-format/wiki/The-Rectangle-Rule)

capstone-frontend/src/components/ClubSearchList.jsx
```diff
@@ -0,0 +1,104 @@
...
+  render() {
+    return (this.state.loading
+      ?
+      (<div className='text-center mt-4'>
+        <Spinner animation='border' role='status' />
+        <br />
+        <h1>Loading...</h1>
+      </div>)
+      :
+      <>
```
geoffbass: It looks to me like this fragment can be removed too, since this subsection resolves to either a `<p>` or a `<CardDeck>`

capstone-frontend/src/components/LeftSideBar.jsx
```diff
@@ -429,7 +429,9 @@ export const LeftSideBar = withRouter((props) => {
           <Col id="main-body">
             <Route exact path='/' component={Home} />
             <Route path='/browse/:query' render={(pageProps) => (
-              <Browse bookLists={props.bookLists} updateBookLists={props.updateBookLists} searchQuery={pageProps.match.params.query} />
+              <Browse bookLists={props.bookLists}
```
geoffbass: nit: move to the line below so it lines up with the other props

geoffbass (changes requested):
Looks pretty good, see if you can move that search logic to the back-end though

capstone-frontend/src/components/BookSearchTile.jsx
antoniolinhart: Good idea. I've changed it based on different mobile sizes and made sure it stays readable and simple across all sizes of screen.

capstone-backend/src/main/java/com/google/sps/servlets/ClubSearchServlet.java
antoniolinhart: Initially I considered doing this before since the doGet is currently 4 lines and I feel like all the "club" related calls should go in the same class and not have separate methods.

I'll move my clubSearch code to the clubServlet to make it more simple.

capstone-backend/src/main/java/com/google/sps/servlets/ClubSearchServlet.java
antoniolinhart: No longer an issue once the clubSearch code was moved to the clubServlet. Steven's Utility code will take care of the error messages if there is no `searchTerm` present.

For the case when there is a non-null searchTerm and more than 1 parameter in the map, I have updated the error message.

capstone-backend/src/main/java/com/google/sps/servlets/ClubSearchServlet.java
antoniolinhart: The ClubSearchServlet class was merged with the ClubServlet and this line has been removed.

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
antoniolinhart: I added local testing lines and accidentally removed the starting line of the Servlet files when I removed them to commit this code.
Fixed.

The formatter we use is Gitpod, but it requires clicking a formatting button and only handles formatting code, it does not deal with extra whitespace like this. If we had the time, I think it would have been a good idea to implement some sort of linter at the beginning of the project to ensure consistency but at this point there is not sufficient time.

capstone-frontend/src/components/Browse.jsx
antoniolinhart: This un-mounting only takes place if the user is trying to navigate away from the search page less than a second after they execute the query. It is uncommon for it to happen where a user is using the application regularly, but I could not figure out a more reliable fix at the moment.

capstone-frontend/src/components/Browse.jsx
antoniolinhart: Yup! Thanks for pointing this out

capstone-frontend/src/components/ClubSearchList.jsx
antoniolinhart: This makes sense! I was running into issues with there being no return value from my fetch when I removed the internal async await.

Met offline to discuss and resolve it and learned more about how Promises work and how to properly return the inside promise.

capstone-frontend/src/components/ClubSearchList.jsx
antoniolinhart: Yup! This is true.

capstone-frontend/src/components/ClubSearchList.jsx
antoniolinhart: I didn't know about this rule before! Thanks! I've indented it another tab to follow accordingly.

capstone-frontend/src/components/Browse.jsx
antoniolinhart: Right! This makes sense, thanks for pointing it out

capstone-frontend/src/components/ClubSearchList.jsx
antoniolinhart: Removed!

capstone-frontend/src/components/ClubSearchList.jsx
antoniolinhart: Discussed offline and decided there is too much refactoring required to be able to make these kind of calls across backend servlets.

Although it is the right choice to do this kind of data joining on the server-side instead of the client-side, our backend is not built for parallel backend calls. Our backend database fetches rely heavily on the request parameters sent from the API call which we execute on the frontend. To make these calls on the backend would require lots of boilerplate code since we cannot modify the Parameter Map once it has already been sent from the frontend. If we could, we would be able to call servlets from other servlets, however refactoring to accommodate for that would take too long at this point, therefore I have decided to keep these fetches on the frontend and fix the await statements so the user does not have to wait longer than necessary.

capstone-frontend/src/components/ClubSearchList.jsx
```diff
@@ -15,60 +15,60 @@ export class ClubSearchList extends Component {
 
   getData = async () => {
     // Get clubs by searchQuery
-    let clubs = await fetch(`/api/clubSearch?searchTerm=${this.props.searchQuery}`)
+    let clubs = await fetch(`/api/clubs?searchTerm=${this.props.searchQuery}`)
```
geoffbass, line 18: nit: Can be `const` now that you're not reassigning

capstone-frontend/src/components/ClubSearchList.jsx
```diff
@@ -15,60 +15,60 @@ export class ClubSearchList extends Component {
...
       });
 
     // Get clubs by checking to see if any clubs are reading the books found in Google Books API search
-    await Promise.all(this.props.books.map(async (book) => {
-      const clubsWithBook = await fetch(`/api/clubs?gbookID=${book.id}`)
+    await Promise.all(this.props.books.map(book => {
+      return fetch(`/api/clubs?gbookID=${book.id}`)
         .then(response => response.json())
+        .then(clubsWithBook => {
+          if (clubsWithBook > 0) {
```
geoffbass: Should this me `clubsWithBook.length`? (assuming this is an array)

capstone-frontend/src/components/ClubSearchList.jsx
```diff
@@ -15,60 +15,60 @@ export class ClubSearchList extends Component {
...
+            clubs.push(clubsWithBook)
+          }
+        })
```
geoffbass: Looks much better!

capstone-frontend/src/components/ClubSearchList.jsx
```diff
@@ -15,60 +15,60 @@ export class ClubSearchList extends Component {
...
 
     // Get clubs by checking to see if any clubs are reading the books found in Google Books API search
-    await Promise.all(this.props.books.map(async (book) => {
-      const clubsWithBook = await fetch(`/api/clubs?gbookID=${book.id}`)
+    await Promise.all(this.props.books.map(book => {
+      return fetch(`/api/clubs?gbookID=${book.id}`)
         .then(response => response.json())
+        .then(clubsWithBook => {
+          if (clubsWithBook > 0) {
+            clubs.push(clubsWithBook)
```
geoffbass: I think if you replace `push` with [`concat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat) here you can omit the later `flat` call.

geoffbass (approved):
LGTM, thanks for making those changes!

capstone-frontend/src/components/ClubSearchList.jsx
antoniolinhart: Yes it should be! Must've gotten missed in the translation

capstone-frontend/src/components/ClubSearchList.jsx
antoniolinhart: This is more simple, thanks!

capstone-frontend/src/components/ClubSearchList.jsx
antoniolinhart: With the new concat instead of push, I have to reassign it with `clubs = clubs.concat(clubsWithBook)` since it does not seem to update the array itself without reassignment.

PR #110, opened 2020-07-28
Search Servlet - Thumbnail Fix

### LINKED ISSUE #72: SearchServlet Thumbnail Retrieval

antoniolinhart:
# Description
Through more testing on our app, I realized that not all books retrieved from the Google Books API are guaranteed to have thumbnails (stored in the imageLinks object). When a search query ran into one of these books, it would crash, and nothing would appear on the frontend.

To avoid this, I added a check to see if the imageLinks object is null. If it is null, then a link to an image depicting a book with a question mark on it will be returned to the frontend. This ensures there is a consistent user experience on the frontend, where they can continue to click on the image and be directed to the book's specific bookpage via the book overlay.

Note: I also added a small fix to the frontend that shows "no rating" when there are no ratings on the book. This was not showing up appropriately before.

# Screenshot of Feature

This is one of the books that lacks a thumbnail, so its image link was replaced with this alternate image.
Unknown Question Mark Book

# Linked Issue
Finished #72

### DIFF
--- a/capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
@@ -225,7 +225,7 @@ public class SearchServlet extends HttpServlet {
       if (greatestSize.isPresent()) {
         thumbnailLink = thumbnailElement.getAsJsonObject().get(greatestSize.get().toString()).getAsString();
       }
-      thumbnailLink = thumbnailLink.replace("http", "https");
+      thumbnailLink = thumbnailLink.replace("http:", "https:");
     }
 
     JsonElement accessInfoElement = bookInfo.get("accessInfo");
--- a/capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
+++ b/capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
@@ -43,6 +43,7 @@ import javax.servlet.http.HttpServletResponse;
 public class SearchServlet extends HttpServlet {
 
   public final static int DEFAULT_MAX_RESULTS = 5;
+  public final static String DEFAULT_THUMBNAIL_IMAGE = "https://upload.wikimedia.org/wikipedia/commons/4/48/Gray_book_question.png";
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
@@ -136,7 +137,8 @@ public class SearchServlet extends HttpServlet {
       }
     }
 
-    Gson gson = new GsonBuilder().disableHtmlEscaping().create();
+    Gson gson = new GsonBuilder().disableHtmlEscaping().setPrettyPrinting().create();
+    response.setCharacterEncoding("UTF8");
     response.getWriter().println(gson.toJson(volumes));
   }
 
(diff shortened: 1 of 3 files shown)


VivekShankar (approved):
LGTM - great job finding and fixing this bug with a nice adjustment on the backend to return a default image!

MauroGuerrero20 (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -216,13 +218,15 @@ static VolumeData individualBookToVolumeData(JsonElement singleBook) {
...
-    if (greatestSize.isPresent()) {
-      thumbnailLink = thumbnailElement.getAsJsonObject().get(greatestSize.get().toString()).getAsString();
+    String thumbnailLink = DEFAULT_THUMBNAIL_IMAGE;
+    if (thumbnailElement != null) {
+      Optional<String> greatestSize = sizes.stream()
+          .filter(size -> thumbnailElement.getAsJsonObject().get(size) != null).findFirst();
+      if (greatestSize.isPresent()) {
+        thumbnailLink = thumbnailElement.getAsJsonObject().get(greatestSize.get().toString()).getAsString();
+      }
+      thumbnailLink = thumbnailLink.replace("http", "https");
```
antoniolinhart: If this somehow executes with greatestSize not being present, there may be an issue replacing the `http` within the DEFAULT_THUMBNAIL_IMAGE where it may end up producing something like `httpss`

steven-solar: Could you instead have `DEFAULT_THUMBNAIL_IMAGE` use `http` to begin with, and move the replace down below the `if` statement? This way regardless of the presence of `greatestSize`, it will always replace `http` with `https`?

capstone-frontend/src/components/BookPage.jsx
```diff
@@ -79,7 +79,7 @@ export class BookPage extends Component {
                         starSpacing='10px'
                         starRatedColor='gold'
                       />
-                      {book.avgRating === 0 && <p id='rating-label'>No Ratings</p>}
+                      {book.avgRating < 1 && <p id='rating-label'>No Ratings</p>}
```
steven-solar, line 82: What was the reasoning for changing this?

steven-solar (approved):
Looks good! Left a comment regarding the http replace issue on the thumbnail link that may fix your problem.

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
geoffbass: Could you avoid this by replacing "http:" with "https:"?

Are all the images available over https as well or does this sometimes result in the default image?

capstone-backend/src/test/java/com/google/sps/servlets/SearchServletTest.java
```diff
@@ -113,7 +113,8 @@ public boolean checkPopulatedVolumeDataObject(VolumeData book) {
   public void ensureJsonResponseIsInCorrectFormat() {
     // Response should be an array of JSON objects, in the form of:
     // [{book1Info}, {book2Info}...]
-
+    // Remove all whitespace as we only care about the beginning and ending characters
+    jsonResponse = jsonResponse.replaceAll("\\s","");
```
geoffbass: Is this actually helpful in the tests if you're only looking at the beginning and ending characters? It's extra code and extra processing so it should really only be included if it has a clear benefit.

capstone-frontend/src/components/BookPage.jsx
geoffbass, line 82: Is it possible for a book to actually have an average rating less than 1 (i.e. can you give it 0 starts or is 1 the lowest)? If not this looks good.

geoffbass (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/servlets/SearchServlet.java
antoniolinhart: Yes Geoff that change would make it so I do not have to worry about this check. I have now changed it to do `http:` to `https:`. I'm not sure why I did not do it this way to start.

I have not run into the case where books are offered only over http and not https. Even when I paste these book links into my browser directly from the API, Google routes me to the https site. This https replace is to avoid the frontend error of serving images from an insecure source.

capstone-frontend/src/components/BookPage.jsx
antoniolinhart: The original goal was for `No Ratings` to pop up next to the stars when there were no ratings found on the backend. On the Google Books site, there is no way to give a book a rating less than 1. On the backend I set the `avgRating` variable to be -1 in the case of no ratings.

With the equality check here for 0, it does not do what I originally intended for it to do, this string did not appear as it should've when there were no ratings.

antoniolinhart: If someone gives a rating to a book on the Google Books site, 1 is the lowest possible rating so there can be no possible average rating below 1.

On the backend I return -1 in the case of no rating because I feel that makes it more clear that it is "invalid" whereas an `avgRating` of 0 could be seen as meaning every person who reviewed gave 0 stars, which is not the case.

capstone-backend/src/test/java/com/google/sps/servlets/SearchServletTest.java
antoniolinhart: This test was originally intended to confirm that the response from the servlet was a proper JSON object. With the change to prettyFormatting (which greatly improves the human readability of the response), the characters were no longer exactly where I expected them to be with the new spacing changes.

Given that the rest of the tests (which are parsing JSON) work based on the JSONResponse being correctly formatted JSON, I think it is reasonable to remove this test since the extra work of the new replaceAll makes the test no longer worth it. I have removed it.

PR #112, opened 2020-07-29
Add Guava Dependency to fix backend breaking

steven-solar:
The backend was not functioning properly due to the absence of the guava dependency in the `pom.xml` file. This PR re-adds this dependency and fixes the backend, unblocking progress on other branches.


antoniolinhart (approved):
LGTM Steven, I looked at the blame and can see that I added it but then it has disappeared since, so I'm not sure what happened.

MauroGuerrero20 (approved):
LGTM

geoffbass (approved):
LGTM

PR #114, opened 2020-07-30
Finish books

steven-solar:
This PR allows users to mark books as done, and then unmark them as done. The `Book` backend object is altered to extend `BaseEntity` to handle `ids` and the `hasRead` field is removed. Book objects will be created when users mark a book as read, and deleted when they unmark as read, so the hasRead field serves no purpose. The `BookServlet` is altered to use the `Utility` class. 

All pages now use the same booktiles, modeled after the inline book tiles from the `ListPage`. All these tiles have a Mark as Read button if a Book object exists, and an Unmark as Read if not. When rendering booktiles, the backend is queried to see if there is a Book object with the given `gbookID` and `userID`. 

Mark as read/unread: 
Screenshot 2020-07-29 at 10 26 40 PM
Screenshot 2020-07-29 at 10 26 10 PM

New search tiles:
Screenshot 2020-07-29 at 10 03 28 PM


capstone-backend/src/main/java/com/google/sps/data/Book.java
```diff
@@ -1,35 +1,27 @@
...
     this.gbookID = gbookID;
-    this.hasRead = hasRead;
     this.rating = rating;
     this.review = review;
   }
  
-  public Book(String id, String userID, String gbookID) {
-    this(id, userID, gbookID, false, -1, "");
+  public Book(String userID, String gbookID) {
+    this(userID, gbookID, -1, "");
```
antoniolinhart, line 20: Even though we're not currently using the rating or review, it's a good idea to continue to include this here for where we are able to implement it in the future!

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -1,34 +1,87 @@
...
+      hasRead: false,
+      bookObject: {},
+    }
+    fetch(`/api/books?userID=${window.localStorage.getItem('userID')}&gbookID=${this.props.book.id}`)
+      .then(response => response.json()).then(books => {
+        if (books.length === 1) {
+          this.setState({hasRead: books.length === 1, bookObject: books[0]});
+        }
+      })
+      .catch(e => console.log(e));
```
antoniolinhart: Why did you choose to do the fetch function here and in the componentDidMount?

If they're both the same - which it looks they are, you could make a function and just call that inside these functions. It would also make it easier to understand what it's doing with a function name tied to the code.

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -1,34 +1,87 @@
...
+
+  markRead = () => {
+    const readBook = {
+      gbookID: this.props.book.id,
+      userID: window.localStorage.getItem('userID'),
+    }
+    fetch('/api/books', {method:'post', body: JSON.stringify(readBook)})
+        .then(response => response.json())
+        .then(bookObject => this.setState({hasRead: true, bookObject}))
+        .catch(e => console.log(e));
```
antoniolinhart, line 26: nit: These catches are console.logs instead of alerts like elsewhere in the application, should we move forward using console logs or alerts?

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -1,34 +1,87 @@
...
+          <br />
+          {this.props.location === 'search' &&
+            <BookListAddDropdown bookLists={this.props.bookLists} updateBookLists={this.props.updateBookLists} book={this.props.book} />}
+          {this.props.location === 'list' &&
+            <Button className="my-4 width-75" variant="danger" onClick={async () => await this.deleteBook(this.props.book.id)}>
+              Remove Book from List
+          </Button>}
+          {this.state.hasRead ? 
+            <Button className='my-4 width-75' variant='danger' onClick={this.markUnread}> Unmark as Read </Button> :
+            <Button className='my-4 width-75' variant='success' onClick={this.markRead}> Mark as Read </Button>}
```
antoniolinhart: Is there a difference between the class w-75 and width-75 or do they both do `width: 75%`?

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -1,34 +1,87 @@
...
+      })
+      .catch(e => console.log(e));
+  }
+
+  render() {
+    return (
+      <Row className="text-center border m-5 bg-light light-gray-border" key={this.props.book.id} >
+        <Col md={3} className="my-4 p-0 ">
+          <BookDescriptionOverlay book={this.props.book}>
+            <img className="img-fluid book-img-md" src={this.props.book.thumbnailLink} alt={this.props.book.title} />
```
antoniolinhart: nit: Mix of classnames in this render function using single and double quotes, you should standardize to single quotes

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -1,34 +1,87 @@
...
+          this.setState({hasRead: books.length === 1, bookObject: books[0]});
+        }
+      })
+      .catch(e => console.log(e));
+  }
+
+  render() {
+    return (
+      <Row className="text-center border m-5 bg-light light-gray-border" key={this.props.book.id} >
+        <Col md={3} className="my-4 p-0 ">
```
antoniolinhart: If the BookSearchTile itself is getting a key when it is created, why does the Row need a key too?

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -1,34 +1,87 @@
...
+          <p className="my-3" id='book-authors'> {this.props.book.authors.join(', ')} </p>
+        </Col>
+
+        <Col md={3} className="my-4 p-0">
+          <a className="btn btn-primary mt-4 width-75" href={this.props.book.webReaderLink}>Web Reader</a>
+          <br />
+          {this.props.location === 'search' &&
+            <BookListAddDropdown bookLists={this.props.bookLists} updateBookLists={this.props.updateBookLists} book={this.props.book} />}
+          {this.props.location === 'list' &&
+            <Button className="my-4 width-75" variant="danger" onClick={async () => await this.deleteBook(this.props.book.id)}>
```
antoniolinhart: Why does this delete need to be awaited? If it does, I feel like it would be clearer to do it in the deleteBook function itself

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -1,34 +1,87 @@
...
+          <h2 className="mt-4" id='book-title'> {this.props.book.title} </h2>
+          <StarRatings
+            rating={this.props.book.avgRating}
+            starDimension="40px"
+            starSpacing="10px"
+            starRatedColor="gold" />
+          <p className="my-3" id='book-authors'> {this.props.book.authors.join(', ')} </p>
+        </Col>
+
+        <Col md={3} className="my-4 p-0">
```
antoniolinhart: Should this size change on different screen sizes?

antoniolinhart (approved):
LGTM.

How does this look on different size screens? I took some time on my current PR to look into what would look the most consistent and clean across sizes.
You don't have to use them, but here are the sizes I came up with. At the smaller sizes I thought the buttons should be all below the image/title because it was too cramped for 3 (out of 12) columns on such a small screen.

For the leftmost thumbnail col: `<Col xs={12} sm={6} md={3}>`
For the middle title/author col:   `<Col xs={12} sm={6} md={5}>`
For the right buttons col:  `<Col xs={12} md={4}>`

In the future it would be cool to break up all the buttons into their own components then bring them into this class as children with props.children where they're necessary so it's more versatile!

capstone-backend/src/main/java/com/google/sps/data/Book.java
steven-solar: Thanks!

capstone-frontend/src/components/BookSearchTile.jsx
steven-solar: I pulled this into a helper function in my new commit!

capstone-frontend/src/components/BookSearchTile.jsx
steven-solar: No difference, will change to `w-75`

capstone-frontend/src/components/BookSearchTile.jsx
steven-solar: sure thing!

capstone-frontend/src/components/BookSearchTile.jsx
steven-solar: This is left over from Mauro's inline implementation, and I forgot to remove it. Thanks for catching.

capstone-frontend/src/components/BookSearchTile.jsx
steven-solar: I don't believe it does, also a remnant from Mauro's original implementation

capstone-frontend/src/components/BookSearchTile.jsx
steven-solar: I believe we should start using console logs, just because some of the error reporting is unhelpful and slows down user experience.

capstone-frontend/src/components/BookSearchTile.jsx
steven-solar: Changing it like you mentioned in your comment.

capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -1,34 +1,85 @@
...
+
+  componentDidMount() {
+    this.fetchBook();
+  }
+
+  fetchBook = () => {
+    fetch(`/api/books?userID=${window.localStorage.getItem('userID')}&gbookID=${this.props.book.id}`)
+      .then(response => response.json()).then(books => {
+        if (books.length === 1) {
+          this.setState({hasRead: books.length === 1, bookObject: books[0]});
```
VivekShankar: nit: Looks like this can just be replaced with `hasRead: True` since you're inside the conditional `if (books.length === 1)`

capstone-backend/src/main/java/com/google/sps/servlets/BookServlet.java
```diff
@@ -47,91 +32,25 @@ public BookServlet() throws IOException {
...
-        if (document.exists()) {
-          book = document.toObject(Book.class);
-        } else {
-          System.err.println("Error: no such document!");
-        }
-      } catch (Exception e) {
-        System.err.println("Error: " + e);
-      }
-      result.add(book);
+    List<Book> retrievedBooks = Utility.get(books, request, response, new GenericClass(Book.class));
```
VivekShankar, line 35: Nice - the methods on this servlet look really clean and concise with the introduction of the `Utility` methods!

VivekShankar (approved):
LGTM

MauroGuerrero20 (approved):
LGTM

capstone-frontend/src/components/BookSearchTile.jsx
steven-solar: Good call, changing this!

capstone-backend/src/main/java/com/google/sps/servlets/BookServlet.java
steven-solar: Thanks!

PR #120, opened 2020-07-31
MyBooks Page

antoniolinhart:
# Description

This PR adds a new MyBooks page to the application, where users can view a list of their read books, in the same form as the BookList or BookSearchList.

When a user clicks that they "read" a book on any other page, they will show up on this page for them to view.

Minor fixes to buttons across the app to maintain the constant width across all buttons.

# Screenshot of Feature

View of my books page with populated books.
My Books Page


capstone-frontend/src/components/BookListAddDropdown.jsx
```diff
@@ -36,8 +36,13 @@ const BookListAddDropdown = ({ book, bookLists, updateBookLists }) => {
     );
   } else {
     return (
-      <CreateList id='button-list-add' selectedBookID={book.id} selectedBook={book} updateBookLists={updateBookLists} className='dropdown-add'
-        btnStyle='btn btn-success btn-margin center-horizontal' />
+      <CreateList
```
steven-solar, line 39: nice styling fix

capstone-frontend/src/components/MyBooks.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+  fetchBooks = () => {
+    fetch(`/api/books?userID=${window.localStorage.getItem('userID')}`)
+      .then(response => response.json())
+      .then(bookObjects => Promise.all(bookObjects.map(book => {
+        return fetch(`/api/search?gbookId=${book.gbookID}`)
+          .then(response => response.json())
+          .then(book => book[0])
+      })))
+      .then(books => this._isMounted && this.setState({ books, loading: false }))
```
steven-solar, line 22: Unfamiliar with `_isMounted` but when I went to read further into it I found this: https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html. Not sure how easy it would be to refactor this, but that link should contain some guidance. If you think it'll take too much time, then not worth doing it now since we're on a time crunch.

steven-solar (approved):
Looks great!

MauroGuerrero20 (approved):
LGTM

capstone-frontend/src/components/MyBooks.jsx
antoniolinhart: I remember talking to Geoff about how to make this fix and he said the same thing about unsubscribing from the callbacks. In this case, there's a lot of boilerplate to add to unsubscribe from these fetches and it may take more time than it's worth.

In the website you linked, it states that `isMounted()` (the function) is bad, but suggests `_isMounted` (the variable) like I have here as an easy migration strategy from `isMounted()`, although not the most optimal.

From the site: `An easy migration strategy for anyone upgrading their code to avoid isMounted() is to track the mounted status yourself. Just set a _isMounted property to true in componentDidMount and set it to false in componentWillUnmount, and use this variable to check your component’s status.`

VivekShankar (approved):
LGTM - this looks great!

capstone-frontend/src/components/MyBooks.jsx
steven-solar: Makes sense, nice job considering time/cleanliness tradeoff. Think this is a good solution!

PR #117, opened 2020-07-30
Home Page

MauroGuerrero20:
# Features
- Creates a Home Page for the website.
  - Includes all of the User's clubs (owner/member)
  - Includes all of the User's booklists (owner/collaborator)

## Minor Features
- Adds Spinners to @steven-solar's ```ClubPage``` and ```MyClubs```.
- Refactor Styling for ClubPage. 
- Viwer mode for SearchUserModal.
  - If the User is a viewer (member/collaborator), then the User can view other Collaborator/Member in the Modal but will not be able to add more. 

# Demo
HomePage Demo

# Demo 2
Landing Page Demo 2

PR #121, opened 2020-07-31
Members Functionality for SearchUserModal

MauroGuerrero20:
# Features
- Adds Members functionality to the ```SearchUserModal```
  - Users can now add members to your club using this modal in the ```ClubPage```
- Removes Friends Functionality from the Modal.

# Demo
SearchUserModal Club Member Demo

### Depends on  https://github.com/STEP-Pod-Capstone-Project/Capstone-Project/pull/117


capstone-backend/src/main/java/com/google/sps/servlets/UserServlet.java
```diff
@@ -69,13 +69,8 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
 
         googleUser = new User(googleId, email, fullname, profileImageUrl,
-            new ImmutableMap.Builder<String, String>()
-              .put("access_token", access_token)
-              .put("idpId", idpId)
-              .put("scope", scope)
-              .put("token_id", tokenId)
-              .put("token_type", token_type)
-              .build());
+            new ImmutableMap.Builder<String, String>().put("access_token", access_token).put("idpId", idpId)
```
steven-solar: `nit`: undo the gitpod format

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -123,6 +123,16 @@ class ClubPage extends Component {
       });
   }
 
+  handleMemberChange = async (member, type) => {
```
steven-solar: why is this async?

steven-solar (changes requested):
In the demo, when you add Vivek as a friend, he appears in the members section of the modal. But, he should only be moved to the bottom if he is added as a club member. Also, does he automatically show up as a club member at the bottom, or does it require a page refresh? We could remove the usercards at the bottom of the club, since they don't serve much purpose now that members can be seen in the modal instead. 

And to make sure, if you are not the club owner then do you still have access to the modal? Members could use the modal just to see club members (no search+add functionality). If they don't have access to the modal, then we can leave usercards at the bottom of the clubpage.

capstone-frontend/src/components/SearchUserModal.jsx
```diff
@@ -45,6 +45,10 @@ export class SearchUserModal extends Component {
       else if (this.props.type === 'booklists' && this.props.bookList) {
         searchResults = searchResults.filter(searchItem => searchItem.id !== window.localStorage.getItem('userID'))
       }
+      // Owner cannot be a Member
+      else if (this.props.type === 'clubs' && this.props.club) {
+        searchResults = searchResults.filter(searchItem => searchItem.id !== window.localStorage.getItem('userID'))
+      }
```
antoniolinhart, line 49: Is there any way to combine this and the previous statement? Maybe with a `||`? The filter line is exactly the same it seems

capstone-frontend/src/components/SearchUserModal.jsx
```diff
@@ -74,6 +78,32 @@ export class SearchUserModal extends Component {
 
   componentDidMount() {
     this.fetchCollaborators();
+    this.fetchFriends();
+
+    // TODO: Update ClubPage based on user input
+
+    if (this.props.club && this.props.type === 'clubs') {
+      if (this.props.club.memberIDs) {
+        if (this.props.club.memberIDs.length > 0) {
```
antoniolinhart: It looks like you already have a check for members > 0 in the fetchMembers function to see if there are no members. Do you need this check here then?

capstone-frontend/src/components/SearchUserModal.jsx
```diff
@@ -100,8 +130,64 @@ export class SearchUserModal extends Component {
...
+    const addedUsersTracker = this.removeDuplicatesArrayJsonId([...this.state.addedUsersTracker, ...members]);
 
+    this.setState({ addedUsersTracker, addedUsers: members });
+  }
+
+  fetchFriends = async () => {
+
+    const userID = window.localStorage.getItem('userID')
+
+    const user = await fetch(`/api/user?id=${userID}`).then(resp => resp.json());
```
antoniolinhart: Should there be a catch here in case the fetch fails?

antoniolinhart (approved):
Looks good to me! This was one of our last core features for the club page and it makes it really clean it use!

I agree with Steven, it doesn't seem to make sense that they get added to the bottom of the modal if they are just added as a friend.

capstone-frontend/src/components/ClubPage.jsx
MauroGuerrero20: Forgot to remove. Thanks

capstone-backend/src/main/java/com/google/sps/servlets/UserServlet.java
MauroGuerrero20: Good catch. This happens once a week.

capstone-frontend/src/components/SearchUserModal.jsx
MauroGuerrero20: I was thinking of combing the conditionals initially, but I feel separating them makes the code more readable. Let me know if I should.

capstone-frontend/src/components/SearchUserModal.jsx
MauroGuerrero20: Yes. Because of ```this.props.club``` can return an empty object ```{}```, thus ```memberIDs``` will be undefined. Consequently, if you try to get the ```.length``` of an undefined variable, React will throw an error and stop the web app.  This is why I have to do multiple checks. 

Note: An empty object ```{}``` is consider to be truthy. ```undefined``` is falsy.

capstone-frontend/src/components/SearchUserModal.jsx
MauroGuerrero20: Ok

MauroGuerrero20:
> Looks good to me! This was one of our last core features for the club page and it makes it really clean it use!
> 
> I agree with Steven, it doesn't seem to make sense that they get added to the bottom of the modal if they are just added as a friend.

Friends and club members will appear at the bottom. This design choice was done because I thought people would like to invite their friends to a club. We should discuss more on a GVC.

steven-solar:
@MauroGuerrero20 I agree friends appearing in the modal would be helpful, but as of now when you add Vivek as a friend he appears under the "Club Members" header which is misleading as he is not a club member. Not sure the best way to display friends in the modal, but maybe another header with your friends beneath it?

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
```diff
@@ -290,6 +290,10 @@ public static JsonObject createRequestBodyJson(HttpServletRequest request) throw
    */
   public static <T extends BaseEntity> T put(CollectionReference collectionReference, HttpServletRequest request, 
       HttpServletResponse response, GenericClass<T> genericClass) throws IOException {
+        response.setHeader("Access-Control-Allow-Methods", "GET");
```
steven-solar: `nit`: remove these lines

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -123,6 +123,25 @@ class ClubPage extends Component {
...
+      club.memberIDs.push(member.id);
+      
+      this.setState({ members: [...this.state.members, member], club });
+    }
+    else if (type === 'remove') {
+
+      const members = this.state.members.filter(knownMember => knownMember.id !== member.id);
+
+      const club = this.state.club;
+      club.memberIDs = club.memberIDs.filter(memberId => memberId !== member.id);
```
steven-solar, line 152: nice fix to the bug we were experiencing!

steven-solar (approved):
Looks great!

capstone-backend/src/main/java/com/google/sps/servlets/Utility.java
MauroGuerrero20: Ok

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -125,6 +125,25 @@ class ClubPage extends Component {
...
+    if (type === 'add') {
+
+      const club = this.state.club;
+      club.memberIDs.push(member.id);
+      
+      this.setState({ members: [...this.state.members, member], club });
+    }
+    else if (type === 'remove') {
+
+      const members = this.state.members.filter(knownMember => knownMember.id !== member.id);
```
VivekShankar: nit: Can you update this line to be within the 80 character limit?

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -125,6 +125,25 @@ class ClubPage extends Component {
...
 
+  handleMemberChange = (member, type) => {
+    if (type === 'add') {
+
+      const club = this.state.club;
+      club.memberIDs.push(member.id);
+      
+      this.setState({ members: [...this.state.members, member], club });
+    }
+    else if (type === 'remove') {
```
VivekShankar: nit: Can you move the `else if` to be on the same line as the final `}` above?

capstone-frontend/src/components/SearchUserModal.jsx
```diff
@@ -67,13 +69,38 @@ export class SearchUserModal extends Component {
...
   }
 
   componentDidMount() {
     this.fetchCollaborators();
+
+    // TODO: Update ClubPage based on user input
+
+    if (this.props.club && this.props.type === 'clubs') {
+      if (this.props.club.memberIDs) {
+        if (this.props.club.memberIDs.length > 0) {
```
VivekShankar: Rather than having three nested conditions here inside of which you call `fetchMembers`, could you just `&&` together all the conditions in a single `if` statement?

capstone-frontend/src/components/SearchUserModal.jsx
```diff
@@ -15,14 +15,12 @@ export class SearchUserModal extends Component {
       typingTimeout: 0,
       searchTerm: '',
       searchResults: [],
-      addedUsersTracker: [],
       addedUsers: [],
```
VivekShankar, line 18: Nice job simplifying the logic here to just using the single `addedUsers` variable to keep track of added club members and added friends!

VivekShankar (approved):
LGTM - this looks great, Mauro!

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -125,6 +125,25 @@ class ClubPage extends Component {
       });
   }
 
+  handleMemberChange = (member, type) => {
+    if (type === 'add') {
+
+      const club = this.state.club;
+      club.memberIDs.push(member.id);
```
geoffbass, line 142: This mutates the current state of the component since `club` is a reference to `this.state.club` (not a copy). You could get around this using [`Object.assign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) or [object spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals).

The difficulty here in properly updating the state highlights the problem of having nested state on a component. You could also consider flattening your state, storing `memberIDs` and any other data from `club` you need access to, rather than the higher-level `club` object.

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -125,6 +125,25 @@ class ClubPage extends Component {
...
+      
+      this.setState({ members: [...this.state.members, member], club });
+    }
+    else if (type === 'remove') {
+
+      const members = this.state.members.filter(knownMember => knownMember.id !== member.id);
+
+      const club = this.state.club;
+      club.memberIDs = club.memberIDs.filter(memberId => memberId !== member.id);
```
geoffbass, line 152: Good use of `filter` to avoid mutating `memberIDs`, but this still mutates `this.state.club`. You could instead make create a `memberIDs` constant here to store the result of the `filter` then set the new state to `club: {...club, memberIDs}`.

capstone-frontend/src/components/SearchUserModal.jsx
```diff
@@ -36,7 +34,7 @@ export class SearchUserModal extends Component {
     else {
       searchResults = await fetch(`/api/userSearch?searchTerm=${searchTerm}`)
         .then(response => response.json())
-        .catch(err => alert(err));
+        .catch(err => console.log(err));
```
geoffbass: nit: `console.error` would probably be better here.

capstone-frontend/src/components/SearchUserModal.jsx
geoffbass: This might be another point in favor of upgrading `memberIDs` to the top-level state in this component

capstone-frontend/src/components/SearchUserModal.jsx
```diff
@@ -67,13 +69,38 @@ export class SearchUserModal extends Component {
...
+          this.fetchMembers();
+        }
+      }
+    }
+  }
+
+  removeDuplicatesArrayJsonId = (array) => {
+
+    let helperArrayIDs = [];
+    let newArray = [];
```
geoffbass: nit: `const` since these aren't being reassigned, you can still mutate them with `push`

capstone-frontend/src/components/SearchUserModal.jsx
```diff
@@ -100,8 +127,34 @@ export class SearchUserModal extends Component {
...
 
+    if (!(this.props.type === 'clubs' && this.props.club)) {
+      return;
+    }
+
+    else if (this.props.club.memberIDs.length === 0) {
+      return;
+    }
+
+    let members = [];
```
geoffbass: nit: `const`

capstone-frontend/src/components/SearchUserModal.jsx
```diff
@@ -100,8 +127,34 @@ export class SearchUserModal extends Component {
...
+
+    await Promise.all(this.props.club.memberIDs.map(async (memberId) => {
+
+      const member = await fetch(`/api/user?id=${memberId}`).then(resp => resp.json());
```
geoffbass: Remove the `async` and `await` here, since that will cause the API requests to go out serially instead of in parallel. The `await Promise.all` above will ensure you don't move on until all the requests have received responses as long as you `return` the result of `fetch` from inside your `map` callback. You can `delete` the `member.tokenObj` inside a `.then` here (or synchronously `map` over the members and do so later), and you can avoid the `push` call if you store the result of `await Promise.all` in `members`.

This is a little tricky: let me know if you'd like to follow up and talk it through together at some point.

capstone-frontend/src/components/SearchUserModal.jsx
```diff
@@ -142,27 +195,43 @@ export class SearchUserModal extends Component {
...
-    if (this.props.type === 'booklists' && this.props.bookList) {
-      this.addUserToBookListCollaborators(this.props.bookList, user);
+    const clubUpdateJson = {
+      id: club.id,
+      add_memberIDs: user.id,
     }
 
-    if (!this.arrayContainsJSONId(this.state.addedUsersTracker, user)) {
-      this.setState({ addedUsersTracker: [...this.state.addedUsersTracker, user] })
+    fetch("/api/clubs", {
```
geoffbass, line 198: nit: Single quotes

geoffbass (approved):
LGTM

capstone-frontend/src/components/ClubPage.jsx
MauroGuerrero20: Ok

capstone-frontend/src/components/ClubPage.jsx
MauroGuerrero20: Ok, I didn't know it made a reference instead of a copy. I will use Object.assign.

capstone-frontend/src/components/ClubPage.jsx
MauroGuerrero20: Ok

capstone-frontend/src/components/SearchUserModal.jsx
MauroGuerrero20: Ok

capstone-frontend/src/components/SearchUserModal.jsx
MauroGuerrero20: Ok.

capstone-frontend/src/components/SearchUserModal.jsx
MauroGuerrero20: Ok

capstone-frontend/src/components/SearchUserModal.jsx
MauroGuerrero20: Ok

capstone-frontend/src/components/SearchUserModal.jsx
MauroGuerrero20: Oh okay, I think I implemented the changes correctly. I will try to change all the Promise.All in the project to speed up the web app.

PR #122, opened 2020-08-01
Event frontend

steven-solar:
This PR renders a club's events as `MeetingCards` on the `ClubPage`. If the current user is the club owner, they can delete meetings (which removes them both from the backend and from everyone's Google Calendars). This is done by saving the `eventID` on event creation, and using that to delete events from the Google Calendar API. Moment.js is used to format meeting start and end DateTimes. 

The owner card is moved to the bottom of the club page. 

Screenshot 2020-08-01 at 4 58 42 PM


MauroGuerrero20 (commented):
Change based branch to ```event_backend``` to remove duplicate changes.

MauroGuerrero20 (approved):
LGTM

capstone-frontend/src/components/MeetingCard.jsx
```diff
@@ -0,0 +1,42 @@
+import React, { Component } from 'react';
+import { Button, Card, Col } from 'react-bootstrap';
+import moment from 'moment';
+
+import '../styles/Groups.css';
+
+
+class MeetingCard extends Component {
```
antoniolinhart: Add `export` here and make this a named import instead of default

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -198,6 +215,10 @@ class ClubPage extends Component {
 
               <div className='text-center'>
                 <div className='description'> {this.state.club.description} </div>
+                <div> Upcoming Meetings: </div>
```
antoniolinhart: Does this have to be a div or can it be a span? Usually divs have more than 1 element inside them to categorize them

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -223,13 +244,14 @@ class ClubPage extends Component {
                   <Button variant='danger' onClick={this.leaveClub}>
                     Leave Club
                   </Button>}
+                <div> Club Members: </div>
```
antoniolinhart: Same as above

antoniolinhart (approved):
LGTM

capstone-backend/src/main/java/com/google/sps/servlets/MeetingServlet.java
```diff
@@ -138,5 +147,14 @@ public void doPut(HttpServletRequest request, HttpServletResponse response) thro
   @Override
   public void doDelete(HttpServletRequest request, HttpServletResponse response) throws IOException {
     Utility.delete(meetings, request, response);
+    JsonObject jsonObject = Utility.createRequestBodyJson(request);
```
geoffbass: nit: `jsonObject` could probably be named after the data it describes, maybe something to do with "calendar"

capstone-frontend/src/components/AdminClubPage.jsx
```diff
@@ -65,13 +65,12 @@ class AdminClubPage extends Component {
     }
     const history = this.props.history;
     fetch(`/api/clubs?id=${this.props.match.params.id}`, { method: 'delete' })
-      .then(function () {
-        history.push('/myclubs');
-      })
+      .then(history.push('/myclubs'))
```
geoffbass: This will cause the navigation to happen immediately after the DELETE request goes out (before you get a response) since it's not inside a function anymore. It might be OK to do this kind of "optimistic" update on the front-end, but you might also need to add some logic to make sure the MyClubs page always has the most up-to-date data (i.e. once the delete has completed, since you could be dealing with a race condition). Otherwise, you can wrap the `history.push` in an arrow function here.

capstone-frontend/src/components/AdminClubPage.jsx
```diff
@@ -83,10 +82,9 @@ class AdminClubPage extends Component {
       endDateTime: new Date(document.getElementById('end-datetime').value).getTime(), 
       attendeeEmails: this.state.members.map(m => m.email),
       organizerEmail: JSON.parse(window.localStorage.getItem('profileObj')).email, 
-    }
+    };
     fetch('/api/meetings', {method: 'post', body: JSON.stringify(meeting)})
-        .then(response => response.json())
-        .then(json => console.log(json))
+        .then(history.push(`/clubpage/${meeting.clubID}`))
```
geoffbass: Same here: this navigation will happen before the POST completes if it isn't wrapped in a function

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -70,6 +72,11 @@ class ClubPage extends Component {
         .then(memberJson => memberJson && this.setState({ members: [...this.state.members, memberJson] }))
         .catch(e => console.log(e));
     }
+
+    fetch(`/api/meetings?clubID=${club.id}`)
+      .then(response => response.json()).then(meetings => this.setState({ meetings }))
```
geoffbass: nit: Move the second `.then` down to the next line for readability.

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -70,6 +72,11 @@ class ClubPage extends Component {
...
+      .catch(e => console.log(e));
```
geoffbass: nit: `console.error` would probably be better here.

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -117,6 +124,16 @@ class ClubPage extends Component {
         .catch(e => console.log(e));
   }
 
+  deleteMeeting = (meetingID) => {
+    let meetings = this.state.meetings;
+    let meetingIDsArray = this.state.meetings.map(m => m.id);
```
geoffbass: nit: `const` for these since they aren't reassigned

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -117,6 +124,16 @@ class ClubPage extends Component {
...
+    const index = meetingIDsArray.indexOf(meetingID);
+    if (index > -1) {
+      meetings.splice(index, 1);
```
geoffbass, line 133: This is actually mutating your component's state, since `meetings` is just a reference to `this.state.meetings` (not a copy). You can fix this by using the `filter` method instead of `splice` (since `splice` alters an array in-place and `filter` returns a new array) or by changing the declaration of `meetings` above to `const meetings = [...this.state.meetings];`

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -198,6 +215,10 @@ class ClubPage extends Component {
 
               <div className='text-center'>
                 <div className='description'> {this.state.club.description} </div>
+                <div> Upcoming Meetings: </div>
+                  <CardDeck className='groups-list-container'>
+                    {this.state.meetings.map(m => <MeetingCard meeting={m} deleteMeeting={this.deleteMeeting} />)}
+                  </CardDeck>
```
geoffbass, line 222: This is really nice and clean, good work!

capstone-frontend/src/components/MeetingCard.jsx
```diff
@@ -0,0 +1,42 @@
...
+  deleteMeeting = () => {
+    const deleteToken = {
+      token: JSON.parse(window.localStorage.getItem('token')),
+      eventID: this.props.meeting.eventID,
+    };
+    fetch(`/api/meetings?id=${this.props.meeting.id}`, { method: 'delete', body: JSON.stringify(deleteToken) })
+      .then(this.props.deleteMeeting(this.props.meeting.id))
```
geoffbass, line 15: Wrap in an arrow function if you want it to depend on a successful DELETE

geoffbass (approved):
LGTM once you address a few nits, most importantly wrapping code in your `.then`s in arrow functions

capstone-frontend/src/components/MeetingCard.jsx
steven-solar: Will do!

capstone-frontend/src/components/ClubPage.jsx
steven-solar: I'll turn it into a span, and give it `display: block` as a style. Thanks for pointing this out!

capstone-backend/src/main/java/com/google/sps/servlets/MeetingServlet.java
steven-solar: Good idea, I'll change this.

capstone-frontend/src/components/ClubPage.jsx
steven-solar: Thanks!

capstone-frontend/src/components/AdminClubPage.jsx
steven-solar: I didn't realize that, thanks for pointing it out. I'll revert this to rely on successful completion before pushing like you mentioned.

capstone-frontend/src/components/ClubPage.jsx
steven-solar: Will do!

capstone-frontend/src/components/ClubPage.jsx
steven-solar: Got it, will fix.

capstone-frontend/src/components/ClubPage.jsx
steven-solar: I didn't realize that! My intention was to avoid manipulating the state directly, since I know that is advised against. I'll change it to use the spread operator like you mentioned so that `meetings` holds a copy of the state, rather than a reference to the state itself. Thanks for the feedback!

capstone-frontend/src/components/ClubPage.jsx
steven-solar: Fixed!

PR #125, opened 2020-08-02
Remove unnecessary alerts, use console logs instead

steven-solar:
Often alerts were "too sensitive" and trigger unnecessarily as state loaded, stopping asynchronous calls. In these cases, console logs would be less obtrusive to user experience and allow the platform to function properly.


antoniolinhart (approved):
Nice change! This should make it easier and smooth to use our app

VivekShankar (approved):
LGTM

MauroGuerrero20 (approved):
LGTM

PR #127, opened 2020-08-03
Fix admin page

steven-solar:
After changing routes in the `LeftSideBar` in a recent PR, the `AdminClubPage` was not appropriately updated to use `this.props.id` instead of `this.props.match.params.id` and was not working. This PR fixes this issue.


VivekShankar (approved):
LGTM

PR #136, opened 2020-08-04
Kick out non-members

steven-solar:
This PR adds a check in `fetchData()` to confirm that you are either the club owner or a member. If not, it alerts the user that they do not have access to the club, and returns them to the `MyClubs` page.

### DIFF
--- a/capstone-frontend/src/components/ClubPage.jsx
+++ b/capstone-frontend/src/components/ClubPage.jsx
@@ -52,8 +52,8 @@ class ClubPage extends Component {
 
     this.setState({ club });
 
-    if (!club.memberIDs.includes(window.localStorage.getItem('userID')) 
-        && club.ownerID !== window.localStorage.getItem('userID')) {
+    if (club.ownerID !== window.localStorage.getItem('userID') 
+        && !club.memberIDs.includes(window.localStorage.getItem('userID'))) {
       alert('You are not a member of this club!');
       this.props.history.push('/myclubs');
     }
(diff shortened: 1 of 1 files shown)


capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -52,6 +52,12 @@ class ClubPage extends Component {
 
     this.setState({ club });
 
+    if (!club.memberIDs.includes(window.localStorage.getItem('userID')) 
+        && club.ownerID !== window.localStorage.getItem('userID')) {
```
antoniolinhart: I think the more expensive check here would be the `.includes` would it be possible to move the owner check to the beginning in place of the member `includes` check. This would help increase short circuiting and remove unnecessary computation

antoniolinhart (approved):
LGTM

capstone-frontend/src/components/ClubPage.jsx
```diff
@@ -52,6 +52,12 @@ class ClubPage extends Component {
...
+      alert('You are not a member of this club!');
```
VivekShankar, line 57: nit: This technically checks that you're neither an owner nor a member of the club, right?

VivekShankar (approved):
LGTM

geoffbass (approved):
LGTM

capstone-frontend/src/components/ClubPage.jsx
steven-solar: Good idea! I think the check is currently fairly quick so this wouldn't be a huge improvement, but I think that it's a good idea and would save more time as club size increases. I'll change this.

capstone-frontend/src/components/ClubPage.jsx
steven-solar: It does! In the case that you're neither a member nor the owner, you won't be able to see any club information. This prevents users from navigating directly to a clubpage link and seeing user comments or any other club information.

PR #147, opened 2020-08-05
Synchronize delete of club members and Possible Login fix

MauroGuerrero20:
# Features/Fixes
- Deleting a club member on the ```ClubPage``` will also update the ```SearchUserModal``` removing the member from ```addedUsers```.

- Possible Login Bug Fix
  - Because of the two distinct but similar project websites:
    - https://sopa-capstone-step-2020.uc.r.appspot.com/
    - https://sopa-capstone-step-2020.appspot.com/
  - A user in theory could be login into both, yet if the user logs out of one, it could cause the user to be unable to logout of the other one. 

  - I think for the future we should try to remove one these and only work on one. Its very confusing and maybe lead to security vulnerabilities. 

# Demo
ClubPage and Modal Synchronous delete Member Demo

### DIFF
--- a/capstone-frontend/src/components/SearchUserModal.jsx
+++ b/capstone-frontend/src/components/SearchUserModal.jsx
@@ -35,32 +35,14 @@ export class SearchUserModal extends Component {
   }
 
   componentDidUpdate(prevProps) {
-    
+
     if (this.props.club && this.props.type === 'clubs') {
       if (this.props.deletedMemberId && (this.props.deletedMemberId !== prevProps.deletedMemberId)) {
 
-        const deleteMemberIndex = this.indexOfMemberArray(this.state.addedUsers, this.props.deletedMemberId);
-
-        if (deleteMemberIndex >= 0) {
-          const addedUsers = [...this.state.addedUsers];
-          addedUsers.splice(deleteMemberIndex, 1);
-
-          this.setState({ addedUsers });
-        }
+        const addedUsers = this.state.addedUsers.filter(addedUser => addedUser.id !== this.props.deletedMemberId);
+        this.setState({ addedUsers });
       }
     }
-  }
-  
-  indexOfMemberArray = (arr, provideMemberId) => {
-    let index = 0
-
-    for (const member of arr) {
-      if (member.id === provideMemberId) {
-        return index;
-      }
-      index += 1;
-    }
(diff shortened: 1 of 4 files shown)


steven-solar (approved):
LGTM

antoniolinhart (approved):
Looks good!

capstone-frontend/src/components/SearchUserModal.jsx
```diff
@@ -34,6 +34,35 @@ export class SearchUserModal extends Component {
...
+  componentDidUpdate(prevProps) {
+    
+    if (this.props.club && this.props.type === 'clubs') {
+      if (this.props.deletedMemberId && (this.props.deletedMemberId !== prevProps.deletedMemberId)) {
+
+        const deleteMemberIndex = this.indexOfMemberArray(this.state.addedUsers, this.props.deletedMemberId);
+
+        if (deleteMemberIndex >= 0) {
+          const addedUsers = [...this.state.addedUsers];
+          addedUsers.splice(deleteMemberIndex, 1);
```
geoffbass: Couldn't you use a `filter` here too?

capstone-frontend/src/components/SearchUserModal.jsx
```diff
@@ -34,6 +34,35 @@ export class SearchUserModal extends Component {
...
+    let index = 0
+
+    for (const member of arr) {
+      if (member.id === provideMemberId) {
+        return index;
+      }
+      index += 1;
+    }
+    return -1
+  }
```
geoffbass: Good to make these sorts of things separate functions like you're doing here, but I think it can be removed if you use `filter` (or possibly `find`?) above

geoffbass (approved):
LGTM, just one suggested change

capstone-frontend/src/components/SearchUserModal.jsx
MauroGuerrero20: Yes, I will refactor

capstone-frontend/src/components/SearchUserModal.jsx
MauroGuerrero20: Ok

MauroGuerrero20:
Ignore fail frontend test. It will be fixed in @steven-solar next PR.

PR #100, opened 2020-07-23
Navbar and LeftSideBar Refactoring

MauroGuerrero20:
# Features
- Utilizes **Material UI** to refactor the styling of the **Navbar** and **LeftSideBar**
- Adds a **RightSideBar** using **material ui**. **Currently commented out**.
- **Fixes some incompatibilities** with @antoniolinhart's pop outs and @steven-solar's clubs
- Nabar and Sidebar now **work with mobile**. Fixes styling issues from before.

# Demo
New Navbar and Sidebar Demo

- On the demo while in mobile debug, the Create List Modal doesn't close the drawer. I have added the ability so it close it automatically.

PR #151, opened 2020-08-07
MyReads Rendering On Unread Book

antoniolinhart:
# Description
This PR ensures that the MyRead page updates when the 'unread book' button is clicked. 

If the current link includes `/myreads` AND the `deleteReadBook` prop is passed, it will call that function, which will filter out the book in the state of MyBooks which matches the id of the removed book.

I originally considered just checking the `/myreads` link, but there are some cases where someone may name their Club or BookList "myreads" which would cause this to fail since the way we calculate links would end up like `/clubs/myreads` which would cause this function to fire incorrectly. For that reason, I ensure that both of them are true.

### DIFF
--- a/capstone-frontend/src/components/BookSearchTile.jsx
+++ b/capstone-frontend/src/components/BookSearchTile.jsx
@@ -28,8 +28,13 @@ class BookSearchTile extends Component {
 
   markUnread = () => {
     fetch(`/api/books?id=${this.state.bookObject.id}`, {method: 'delete'})
-        .then(this.setState({hasRead: false, bookObject: {}}))
-        .catch(e => console.log(e));
+        .then(() => this.setState({hasRead: false, bookObject: {}}))
+        .catch(e => console.error(e));
+
+    // Remove book from MyBooks state if on that page
+    if (this.props.deleteReadBook && window.location.pathname.includes('/myreads')) {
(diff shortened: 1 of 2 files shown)


capstone-frontend/src/components/BookSearchTile.jsx
```diff
@@ -28,8 +28,13 @@ class BookSearchTile extends Component {
 
   markUnread = () => {
     fetch(`/api/books?id=${this.state.bookObject.id}`, {method: 'delete'})
-        .then(this.setState({hasRead: false, bookObject: {}}))
-        .catch(e => console.log(e));
+        .then(() => this.setState({hasRead: false, bookObject: {}}))
```
steven-solar, line 31: nice job removing "optimistic behavior" in your promise chaining!

steven-solar (approved):
#100!!!

capstone-frontend/src/components/BookSearchTile.jsx
antoniolinhart: Thanks! I realized this when I was going back to change this
I didn't know that was the term for it!

MauroGuerrero20 (approved):
LGTM

VivekShankar (approved):
LGTM - nice catch!!

PR #154, opened 2020-08-08
Update User Token Security

MauroGuerrero20:
# Features
- Important User token information removed from the Frontend
  - ```localStorage``` now only contains the expiration date of the token
- Calendar Meetings Backend can fetch the token object from Firestore

## Important Note:
@steven-solar Can you make sure I didn't break your calendar meetings feature. It works for me when I tested it.

My changes are currently deployed so you can easily test it.

If you get login issues, you probably should first clear the local storage of the application first.


capstone-backend/src/main/java/com/google/sps/servlets/MeetingServlet.java
```diff
@@ -66,7 +67,34 @@ public Calendar createCalendar(JsonObject jsonObject) {
...
+      StringBuilder tokenString = new StringBuilder(user.get("tokenObj").toString());
+      tokenString.deleteCharAt(0);
+      tokenString.deleteCharAt(tokenString.length() - 1);
+
+      Map<String, String> tokenMap = Arrays.stream(tokenString.toString().split(",")).map(s -> s.split("="))
+          .collect(Collectors.toMap(s -> s[0].replaceAll("\\s", ""), s -> s[1]));
+
+      for (String key : tokenMap.keySet()) {
+        tokenObject.addProperty(key, tokenMap.get(key));
+      }
```
antoniolinhart, line 90: Why are you parsing it like this instead of Gson parse on the tokenObj? It looks like you are just manually parsing the Gson instead of the library

MauroGuerrero20: Hi Antonio, I initially tried to parse it using ```gson``` but the way Firestore sends and formats the user's ```tokenObj``` is incompatible with ```Json.parser```.

antoniolinhart (approved):
LGTM - I'll leave it to steven to ensure calendar functionality is fine

